PR #7, opened 2020-02-20
Create lint and test workflow in github actions

wfvining:
Based on the pythonpackage template

### DIFF
--- a/.github/workflows/lint-and-test.yml
+++ b/.github/workflows/lint-and-test.yml
@@ -25,10 +25,7 @@ jobs:
     - name: Lint with flake8
       run: |
         pip install flake8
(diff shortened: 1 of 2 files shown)


wfvining:
Looks like github actions will work, just fails because there are no tests yet (pytest considers this an error)

wholmgren:
Looks cool. I'm definitely interested in seeing it in action as you add real code and tests.

CameronTStark:
This looks great! Makes me want to play with the GitHub actions once I settle down with our Azure setup. 

Just as a note, it's typical to separate the linting into it's own "step" and do it only for one OS/Python version. Linting isn't resource intensive but it helps highlight it's results if it's on its own.

PR #10, opened 2020-02-27
Configure setuptools

wfvining:
Adds a `setup.py` for packaging and publishing the library.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -9,7 +9,7 @@ DESCRIPTION = ('PVAnalytics is a python library for the analysis of ' +
                'photovoltaic system-level data.')
 
 LONG_DESCRIPTION = """
-PVAnalytics is a collection point for funcions for working with data
+PVAnalytics is a collection of funcions for working with data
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -0,0 +1,75 @@
...
+try:
+    from setuptools import setup
+except ImportError:
+    raise RuntimeError('setuptools is required')
+
+DESCRIPTION = ('PVAnalytics is a python library for the analysis of ' +
+               'photovoltaic system-level data.')
+
+LONG_DESCRIPTION = """
+PVAnalytics is a collection of funcions for working with data
```
wfvining:
```suggestion
PVAnalytics is a collection of functions for working with data
```

setup.py
```diff
@@ -0,0 +1,75 @@
...
+from photovoltaic power systems. The library includes functions for
+general data quality tests such as outlier detection, validataion that
```
wfvining:
```suggestion
general data quality tests such as outlier detection, validation that
```

PR #19, opened 2020-03-05
Implementation of the QCRad algorithm for irradiance quality control

wfvining:
Ports the implementation of the QCRad algorithm from [SolarArbiter/solarforecastarbiter-core](https://github.com/SolarArbiter/solarforecastarbiter-core)

## Changes from SolarForecastArbiter

* removes the `mask_flags` decorator
* don't set `Series.name` for returned masks
* Updated documentation for functions. Mostly rewording, but also refine focus/meaning in some places


pvanalytics/quality/irradiance.py
```diff
@@ -1 +1,333 @@
...
+def _qcrad_ub(dni_extra, sza, lim):
+    cosd_sza = cosd(sza)
+    cosd_sza[cosd_sza < 0] = 0
+    return lim['mult'] * dni_extra * cosd_sza**lim['exp'] + lim['min']
+
+
+def check_ghi_limits_qcrad(ghi, solar_zenith, dni_extra, limits=None):
+    r"""Test for physical limits on GHI using the QCRad criteria.
+
+    Test passes if a value > lower bound and value < upper bound. Lower bounds
```
cwhanse:
```suggestion
    Test is applied to each GHI value. A GHI value passes value > lower bound and value < upper bound. Lower bounds
```

pvanalytics/quality/irradiance.py
```diff
@@ -1 +1,333 @@
...
+    dni_extra : Series
+        Extraterrestrial normal irradiance in :math:`W/m^2`
+    limits : dict, default QCRAD_LIMITS
+        for keys 'ghi_ub', 'dhi_ub', 'dni_ub', value is a dict with
+        keys {'mult', 'exp', 'min'}. For keys 'ghi_lb', 'dhi_lb', 'dni_lb',
+        value is a float.
+
+    Returns
+    -------
+    ghi_limit_flag : Series
```
cwhanse:
```suggestion
    Series
```

pvanalytics/quality/irradiance.py
```diff
@@ -1 +1,333 @@
...
+        True if value passes physically-possible test
```
cwhanse:
```suggestion
        True where value passes limits test
```

pvanalytics/quality/irradiance.py
```diff
@@ -1 +1,333 @@
...
+    ----------
+    ghi : Series
+        Global horizontal irradiance in :math:`W/m^2`
+    solar_zenith : Series
+        Solar zenith angle in degrees
+    dni_extra : Series
+        Extraterrestrial normal irradiance in :math:`W/m^2`
+    limits : dict, default QCRAD_LIMITS
+        for keys 'ghi_ub', 'dhi_ub', 'dni_ub', value is a dict with
+        keys {'mult', 'exp', 'min'}. For keys 'ghi_lb', 'dhi_lb', 'dni_lb',
```
cwhanse:
```suggestion
        keys {'mult', 'exp', 'min'} and float values. For keys 'ghi_lb', 'dhi_lb', 'dni_lb',
```

pvanalytics/quality/irradiance.py
```diff
@@ -1 +1,333 @@
...
+
+    Parameters
+    ----------
+    ghi : Series
+        Global horizontal irradiance in :math:`W/m^2`
+    solar_zenith : Series
+        Solar zenith angle in degrees
+    dni_extra : Series
+        Extraterrestrial normal irradiance in :math:`W/m^2`
+    limits : dict, default QCRAD_LIMITS
```
cwhanse: This function tests only GHI, should we instead describe a dict with the `ghi_ub` and `ghi_lb` keys? Could still be QCRAD_LIMITS by default.

pvanalytics/quality/irradiance.py
```diff
@@ -1 +1,333 @@
...
+    ghi_limit_flag : Series
+        True if value passes physically-possible test
+
+    """
+    if not limits:
+        limits = QCRAD_LIMITS
+    ghi_ub = _qcrad_ub(dni_extra, solar_zenith, limits['ghi_ub'])
+
+    ghi_limit_flag = _check_limits(ghi, limits['ghi_lb'], ghi_ub)
+    ghi_limit_flag.name = 'ghi_limit_flag'
```
cwhanse: I'm not sure how valuable the `Series.name` attribute is.

pvanalytics/quality/irradiance.py
```diff
@@ -1 +1,333 @@
...
+    dni_limit_flag.name = 'dni_limit_flag'
+
+    return dni_limit_flag
+
+
+def check_irradiance_limits_qcrad(solar_zenith, dni_extra, ghi=None, dhi=None,
+                                  dni=None, limits=None):
+    r"""Test for physical limits on GHI, DHI or DNI using the QCRad criteria.
+
+    Criteria from [1]_ are used to determine lower and upper bounds
```
cwhanse:
```suggestion
    Criteria from [1]_ are used to determine physically plausible lower and upper bounds.
```

pvanalytics/quality/irradiance.py
```diff
@@ -1 +1,333 @@
...
+    for physically plausible value. Test passes if a value > lower
```
cwhanse:
```suggestion
    Each value is tested, and a value passes if value > lower
```

wfvining:
Yes, the tests are also from solarforecastarbiter.

pvanalytics/quality/irradiance.py
wfvining: I don't think it is particularly valuable in this context. I will remove the names. (I'll also need to modify the tests to pass `check_names=False` to `assert_series_equal`, since I believe the column name is used as the series name when you get a series out of a dataframe).

pvanalytics/quality/irradiance.py
```diff
@@ -1 +1,333 @@
...
+        raise ValueError('must provide either upper or lower bound')
+
+
+def _qcrad_ub(dni_extra, sza, lim):
+    cosd_sza = cosd(sza)
+    cosd_sza[cosd_sza < 0] = 0
+    return lim['mult'] * dni_extra * cosd_sza**lim['exp'] + lim['min']
+
+
+def check_ghi_limits_qcrad(ghi, solar_zenith, dni_extra, limits=None):
```
wfvining: Is it better style to do:
```python
def check_ghi_limits_qcrad(ghi, solar_zenith, dni_extra, limits=QCRAD_LIMITS):
```
and remove the 
```python
    if not limits:
        limits = QCRAD_LIMITS
```
block below?

pvanalytics/quality/irradiance.py
wfvining: This seems reasonable. We should do the same for `check_[dni/dhi]_limits`

pvanalytics/quality/irradiance.py
wfvining: The same pattern is used in `check_[dni/dhi]_limits

cwhanse: The second pattern is better, since `dict` is mutable, it's discouraged to use as a default value (could be changed inside the function and the change persists outside).

pvanalytics/quality/irradiance.py
wfvining: Do you think we should remove all the return value names? I think they don't provide much information for functions that return a single value, but are informative for functions that return tuples.

pvanalytics/quality/irradiance.py
wfvining: Will take care of this when rewriting description of `limits` param

pvanalytics/quality/irradiance.py
cwhanse: I do, if a function has a single return variable, rather than name it, we could just provide the type.

pvanalytics/quality/irradiance.py
wholmgren: This change make sense here in a general purpose library so that you can minimize the surface of your API. But, for context, these names are used in SFA to simplify the process of combining many series into a dataframe. If that kind of combine many masks together code is in scope of `pvanalytics` then you may want to rethink the change.

wholmgren:
A general request as you port algorithms from SFA to pvanalytics... I'd like to think that SFA will eventually use the implementations here, and it would be helpful if the PRs summarized any changes made in the process. Perhaps the initial PR comment could be edited to keep track of the changes.

pvanalytics/quality/irradiance.py
cwhanse: That's a very good point @wholmgren I was not aware of how the Series names were being used.

wfvining: I think at this point in development we should aim to minimize API surface. I'm further inclined to remove the names because it is difficult to know what a "good" name should be. In particular, if we end up with several different algorithms for generating a particular mask should the name include an identifier for the algorithm that was used? (I'm thinking of inverter clipping especially - I have seen three different methods for detecting this feature in different places and we may want to include several options in pvanalytics.)

If we return named series we would need to require that different implementations of the same filter return the same name or all implementations return different names (ie. by appending `_qcrad` to the name). I think both alternatives are error-prone and introduce unnecessary technical debt.

I think I see a few paths forward (but there are probably more):
* Leave it to the users who need names to add them.
* Defer naming series until a higher-level API exists for pvanalytics and add names at that level. (I'm thinking of the API for pvlib which uses a keyword argument to access different implementations.)
* Add a keyword arg to the filtering/masking functions that allows users to specify a name (but defaults to `None`). For the most part this could be accomplished with minimal changes via a decorator, but there are some filtering functions that return tuples which could complicate the decorator approach.

I prefer the first option, but if it will make the library more difficult to use then we shouldn't do it. I don't think it adds much overhead for SFA since I believe you will need to define wrapper functions anyway to apply the `mask_flags` decorator.

@wholmgren, I really appreciate your insight as a potential future user of this library.

PR #21, opened 2020-03-26
Outlier detection functions

wfvining:
Implements two functions for identifying outliers

* `quality.outliers.tukey()` which uses the interquartile range for outlier deteciton
* `quality.outliers.zscore()` which uses the z-score for outlier detection


dirkjordan:
@wfvining  Thanks, these outlier filters are a good start and perform reasonably well for roughly symmetrically distributed data but are limited for highly skewed data. Unfortunately, outdoor PV data are in that category of highly skewed distributions. I would suggest adding some more robust outlier detectors, like the Hampel filter. It is based on the median and the median absolute deviation. You can also improve it by not using the medcouple instead of the median.

pvanalytics/quality/outliers.py
```diff
@@ -1 +1,62 @@
 """Functions for identifying and labeling outliers."""
+import pandas as pd
+from scipy import stats
+
+
+def tukey(data, k=1.5):
+    r"""Identify outliers based on the interquartile range.
+
+    A value is considered an outlier if it does *not* satisfy the
```
cwhanse:
```suggestion
    A value `x` is considered an outlier if it does *not* satisfy the
```

pvanalytics/quality/outliers.py
```diff
@@ -1 +1,62 @@
...
+
+    where :math:`Q_1` is the value of the first quartile and
+    :math:`Q_3` is the value of the third quartile.
+
+    Parameters
+    ----------
+    data : Series
+        The data in which to find outliers.
+    k : float, default 1.5
+        Multiplier of the IQR. Using a larger value will be more
```
cwhanse:
```suggestion
        Multiplier of the interquartile range. A larger value will be more
```

pvanalytics/quality/outliers.py
```diff
@@ -1 +1,62 @@
...
+        permissive of values that are far from the median.
+
+    Returns
+    -------
+    Series
+        A series of booleans with True for each value that is not an
```
cwhanse: Should the function maintain the grammar, for readability? As written, `quality.outliers.tukey` returns `True` for outliers, which means that the code using the function would read

```
outlier = quality.outliers.tukey(x)
if not outlier:
   do stuff with outliers
```

pvanalytics/quality/outliers.py
```diff
@@ -1 +1,62 @@
...
+    """
+    first_quartile = data.quantile(0.25)
+    third_quartile = data.quantile(0.75)
+    iqr = third_quartile - first_quartile
+    return ~((data < (first_quartile - k*iqr))
+             | (data > (third_quartile + k*iqr)))
+
+
+def zscore(data, zmax=1.5):
+    """Identify outliers based on the z-score.
```
cwhanse:
```suggestion
    """Identify outliers using the z-score.
```

pvanalytics/quality/outliers.py
```diff
@@ -1 +1,62 @@
...
+
+    If the absolute value of the z-score is greater than `zmax` then
```
cwhanse:
```suggestion
    Points with z-score greater than `zmax` are considered as outliers.
```

pvanalytics/quality/outliers.py
wfvining: As written this returns `True` for any value that is _not_ an outlier. I wrote it in this way to make it easy to exclude outliers:

```python
no_outliers = data[outliers.tukey(data)]
```

Is your suggestion to return `False` for any value that is not an outlier and `True` for outliers?

```python
no_outliers = data[~outliers.tukey(data)]
```

Now that I write those examples out (along with your example above) the latter seems like a better choice.

wfvining:
Closes #23 

Adds an implementation of the outlier detection using the Hampel identifier.

pvanalytics/quality/outliers.py
```diff
@@ -1 +1,101 @@
...
+    iqr = third_quartile - first_quartile
+    return ((data < (first_quartile - k*iqr))
+            | (data > (third_quartile + k*iqr)))
+
+
+def zscore(data, zmax=1.5):
+    """Identify outliers using the z-score.
+
+    Points with z-score greater than `zmax` are considered as outliers.
+    the value is considered an outlier.
```
cwhanse, line 44: leftover text here I think

pvanalytics/quality/outliers.py
```diff
@@ -1 +1,101 @@
...
+    r"""Identify outliers by the Hampel identifier.
+
+    The Hampel identifier is computed according to [1]_.
+
+    Parameters
+    ----------
+    data : Series
+        The data in which to find outliers.
+    window : int or offset, default 5
+        The size of the rolling window used to compute the hampel
```
cwhanse, line 73:
```suggestion
        The size of the rolling window used to compute the Hampel
```

requirements.txt
```diff
@@ -1,3 +1,4 @@
 numpy>=1.16.0
 pandas>=0.25.0
 pvlib>=0.7.0
+scipy>=1.4.0
```
cwhanse, line 4: is scipy>=1.4 required for the functions being used here?

wfvining: `scipy.stats.median_absolute_deviation` was added in 1.3. I can reduce the requirement to that.

PR #22, opened 2020-03-26
Port interpolation and stale value detection from SolarForecastArbiter

wfvining:
Adds a `gaps` submodule to the `quality` module for functions that identify gaps in the data that have been filled in (either by interpolation or "stale values").

## Changes from SolarForecastArbiter

* remove `@mask_flags` decorator
* rename `detect_stale_values` to `stale_values_diff`
* rename `detect_interpolation` to `interpolation_diff`


wfvining:
I am considering changing the names of these two functions to `stale_values` and `interpolation`. I like the succinct and declarative sound of these names better for `pvanalytics`.

```python
from pvanalytics.quality import gaps
# ...
suspect_data = data[gaps.stale_values(data) | gaps.interpolation(data)]
```
vs.
```python
from pvanalytics.quality import gaps
# ...
suspect_data = data[gaps.detect_stale_values(data) | gaps.detect_interpolation(data)]
```

This matches the naming style we are planning on using in the outliers module (#21), but is different than the naming used for the qcrad functions. We probably want to commit to a consistent naming convention for these functions.

wholmgren:
SFA issue to be aware of, though I can't recommend a specific solution here: https://github.com/SolarArbiter/solarforecastarbiter-core/issues/124

cwhanse:
`interpolated` rather than `interpolation`? 

As @wholmgren noted, the algorithm can be fooled by data that has been rounded or truncated. We should anticipate other functions that detect stale or interpolated values. The pvlib-like pattern would be `pvanalytics.gaps.detect_stale_XXX` where XXX is the source of the algorithm. In this case, there's no publication to reference. Perhaps `sfa` would be appropriate.

wholmgren:
> Perhaps sfa would be appropriate.

I don't think the Arbiter needs credit for these functions, especially since you contributed them. If Arbiter had followed the pattern then it might be known as `detect_stale_sandia` or `_hansen`.

cwhanse:
> I don't think the Arbiter needs credit for these functions, especially since you contributed them. If Arbiter had followed the pattern then it might be known as detect_stale_sandia or _hansen.

Credit or blame... `_sandia` is OK.

wfvining:
>  We should anticipate other functions that detect stale or interpolated values. The pvlib-like pattern would be `pvanalytics.gaps.detect_stale_XXX` where XXX is the source of the algorithm. In this case, there's no publication to reference. Perhaps `sfa` would be appropriate.

How about `_delta` or `_diff`? I'm imagining of other methods based off of things like the goodness of a linear fit in a window (`_fit`) or curvature (`_curvature`)

cwhanse:
> How about _delta or _diff? I'm imagining of other methods based off of things like the goodness of a linear fit in a window (_fit) or curvature (_curvature)

Either `_delta` or `_diff` is OK also.

PR #27, opened 2020-04-02
Port weather data validation checks from SolarForecastArbiter

wfvining:
Extreme value filters for air temperature, relative humidity, and wind speed ported from SolarForecastArbiter (SFA). These functions just wrap `quality.util.check_limits()` providing reasonable defaults for each type of data.

## Changes from SFA

* remove `@mask_flags` decorator
* the name of the returned Boolean series is not set.
* rename `validator.check_temperature_limits()` to `weather.temperature_limits()`
* rename `validator.check_rh_limits()` to `weather.relative_humidity_limits()`
* rename `validator.check_wind_limits()` to `weather.wind_limits()`

### DIFF
--- a/pvanalytics/quality/weather.py
+++ b/pvanalytics/quality/weather.py
@@ -22,3 +22,29 @@ def temperature_limits(air_temperature, limits=(-35.0, 50.0)):
     return util.check_limits(
         air_temperature, lower_bound=limits[0], upper_bound=limits[1]
     )
+
+
+def relative_humidity_limits(relative_humidity, limits=(0, 100)):
+    """Check for extremes in relative humidity data.
+
+    Parameters
+    ----------
+    relative_humidity : Series
+        Relative humidity in %.
+    limits : tuple, default (0, 100)
+        (lower bound, upper bound) for relative humidity.
+
+    Returns
+    -------
+    Series
+        True if `relative_humidity` >= lower bound and
+        `relative_humidity` <= upper_bound.
+
+    """
+    return util.check_limits(
+        relative_humidity,
+        lower_bound=limits[0],
+        upper_bound=limits[1],
+        inclusive_lower=True,
+        inclusive_upper=True
+    )
--- a/pvanalytics/quality/weather.py
+++ b/pvanalytics/quality/weather.py
@@ -1,1 +1,24 @@
 """Quality control functions for weather data."""
+from pvanalytics.quality import util
+
+
+def temperature_limits(air_temperature, limits=(-35.0, 50.0)):
+    """Identify extreme temperatures.
+
+    Parameters
+    ----------
+    air_temperature : Series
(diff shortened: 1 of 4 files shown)


pvanalytics/quality/weather.py
```diff
@@ -1 +1,75 @@
 """Quality control functions for weather data."""
+from pvanalytics.quality import util
+
+
+def temperature_limits(air_temperature, limits=(-35.0, 50.0)):
+    """Identify extreme temperatures.
```
cwhanse, line 6:
```suggestion
    """Identify temperature values that are within limits.
```

pvanalytics/quality/weather.py
```diff
@@ -1 +1,75 @@
...
+
+    Parameters
+    ----------
+    air_temperature : Series
+        Air temperature in Celsius.
```
cwhanse, line 11: For discussion: do we want to use RTD auto-format for units?  It would be 
```suggestion
        Air temperature [C].
```

pvanalytics/quality/weather.py
```diff
@@ -1 +1,75 @@
...
+        < upper bound.
+
+    """
+    return util.check_limits(
+        air_temperature, lower_bound=limits[0], upper_bound=limits[1]
+    )
+
+
+def relative_humidity_limits(relative_humidity, limits=(0, 100)):
+    """Check for extremes in relative humidity data.
```
cwhanse, line 28:
```suggestion
    """Identify relative humidity values that are within limits.
```

pvanalytics/quality/weather.py
```diff
@@ -1 +1,75 @@
...
+        relative_humidity,
+        lower_bound=limits[0],
+        upper_bound=limits[1],
+        inclusive_lower=True,
+        inclusive_upper=True
+    )
+
+
+def wind_limits(wind_speed, limits=(0.0, 50.0)):
+    """Check for extreme wind speeds.
```
cwhanse, line 54:
```suggestion
    """Identify wind speed values that are within limits.
```

pvanalytics/quality/weather.py
wfvining: This seems like a good idea. Can you point me towards some documentation for this feature? In other parts of the code I have been using something like `` :math:`W/m^2` `` to format units like 'W/m^2', but if possible I would much prefer [W/m^2]

cwhanse: So, looking at it again, I don't think it's actually formatting anything, just putting text braces around the unit. pvlib tries to stick with SI unit symbols. One [example ](https://pvlib-python.readthedocs.io/en/stable/generated/pvlib.temperature.sapm_cell.html) docstring using the braces. I think you could [`:math:`W/m^2`] if we want to format with superscripts.

PR #28, opened 2020-04-02
Port timestamp spacing check from SolarForecastArbiter

wfvining:
Validate that timestamp spacing matches the expected spacing.

## Changes from SFA

* rename `check_timestamp_spacing()` to `quality.time.spacing()`
* remove the `@mask_flags` decorator
* Don't pass `keep_tz` to `DatetimeIndex.to_series()`. This keyword argument was deprecated in pandas 0.24.0. Since pvanalytics requires pandas >= 0.25.0 we remove it.


wfvining:
@cwhanse I think this function provides a basis for that capability (`not quality.time.spacing(...).all()` tells you if there are missing timestamps or non-uniformities). Other capabilities are easily built from this as well. For example, get the longest span of uniformly spaced time-stamps to use for detecting clearsky conditions (or other operations that are sensitive to missing/irregular time stamps).

cwhanse:
+1 to merging this

PR #30, opened 2020-04-07
Port clearsky-based quality checks from Solar Forecast Arbiter

wfvining:
Adds quality test to flag irradiance measurements that exceed the expected clearsky irradiance.

## Changes from SFA
* consolidate POA and GHI checks into a single function. These were separate in SFA, but ultimately do exactly the same thing. For pvanalytics it makes sense to consolidate them into a single general purpose function: `quality.irradiance.clearsky_limits()`.
* remove the `@mask_flags` decorator.

### DIFF
--- a/pvanalytics/quality/irradiance.py
+++ b/pvanalytics/quality/irradiance.py
@@ -313,12 +313,12 @@ def check_irradiance_consistency_qcrad(ghi, solar_zenith, dhi, dni,
 
 
 def clearsky_limits(measured, clearsky, csi_max=1.1):
-    """Identify irradiance measurements greater than clearsky values.
+    """Identify irradiance values which do not exceed clearsky values.
 
     Uses :py:func:`pvlib.irradiance.clearsky_index` to compute the
-    clearsky index for `measured` and `clearsky`. Compares the
+    clearsky index as the ratio of `measured` to `clearsky`. Compares the
     clearsky index to `csi_max` to identify values in `measured` that
-    are greater than the expected clearsky irradiance.
+    are less than or equal to `csi_max`.
 
     Parameters
     ----------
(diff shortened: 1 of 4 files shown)


pvanalytics/quality/irradiance.py
```diff
@@ -309,3 +310,35 @@ def check_irradiance_consistency_qcrad(ghi, solar_zenith, dhi, dni,
                              bounds=bounds['low_zenith']))
 
     return consistent_components, diffuse_ratio_limit
+
+
+def clearsky_limits(measured, clearsky, csi_max=1.1):
+    """Identify irradiance measurements greater than clearsky values.
```
cwhanse, line 316: Here, we "identify...greater than"  The function returns True when a value is less than or equal to.  Do we want to return True if greater than, or keep this behavior?

wfvining: I think we want to keep the current behavior. It matches the semantics of the qcrad validation functions in this module which return True for values that are physically plausible.

To make the documentation more clear, perhaps it should say "Identify irradiance measurements that do not exceed clearsky values"

cwhanse, line 316:
```suggestion
    """Identify irradiance values which do not exceed clearsky values.
```

pvanalytics/quality/irradiance.py
```diff
@@ -309,3 +310,35 @@ def check_irradiance_consistency_qcrad(ghi, solar_zenith, dhi, dni,
...
+
+    Uses :py:func:`pvlib.irradiance.clearsky_index` to compute the
+    clearsky index for `measured` and `clearsky`. Compares the
```
cwhanse, line 319:
```suggestion
    clearsky index as the ratio of `measured` to `clearsky`. Compares the
```

pvanalytics/quality/irradiance.py
```diff
@@ -309,3 +310,35 @@ def check_irradiance_consistency_qcrad(ghi, solar_zenith, dhi, dni,
...
+    clearsky index to `csi_max` to identify values in `measured` that
+    are greater than the expected clearsky irradiance.
```
cwhanse, line 321:
```suggestion
    are less than or equal to `csi_max`.
```

PR #32, opened 2020-04-16
Port detect_clearsky from Solar Forecast Arbiter

wfvining:
## Changes from SFA
* renamed `validator.detect_clearsky` to `features.clearsky.reno`
* wrote new tests that do not depend on pvlib to generate cleasky data.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -103,6 +103,14 @@ Functions for identifying inverter clipping
 
    features.clipping.levels
 
+Clearsky
+--------
+
+.. autosummary::
+   :toctree: generated/
+
+   features.clearsky.reno
+
 .. rubric:: References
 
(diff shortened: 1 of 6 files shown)


wholmgren:
I’d put it in pvlib. No reason to make users install 2 libraries at this point. I’d reconsider if/when this library’s features and scope are better understood and adopted.

wfvining:
I certainly prefer the submodule approach for naming these functions (`features.clearsky.reno()`). If we take the submodule route, should we make a submodule for each feature (i.e. `features.clipping` with a `levels()` function in it) to keep things consistent?

cwhanse:
>  If we take the submodule route, should we make a submodule for each feature (i.e. features.clipping with a levels() function in it) to keep things consistent?

Yes.  We should create space for other clipping methods to be added, e.g., the 99.5th percentile approach being used for RdTools.

PR #34, opened 2020-04-21
Add copyright notice to each function ported from SolarForecastArbiter

wfvining:
Previously the required copyright notice was maintained in the documentation tree and a list of modules/functions to which it applied was kept with the license.

This PR adds finer-grained copyright notices which will be helpful as we incorporate functionality from other projects. Keeping the copyright in the individual function's docstring should make tracking/updating copyright notices simpler in the future.

To see how the docs look with the attribution in-line see https://pvanalytics.readthedocs.io/en/per-function-attribution/


wholmgren:
FWIW I thought it's ok to just put `# Copyright Org` in the source code rather than function doc strings (while retaining the full license somewhere else). Wondering if you're following a pattern that I'm not aware of.

wfvining:
@wholmgren I was following a pattern from the Software Freedom Law Center here: http://softwarefreedom.org/resources/2012/ManagingCopyrightInformation.html#maintaining-centralized-copyright-notices

It is a bit verbose, I agree.

wfvining:
Sorry, didn't read your full comment... I prefer docstrings because they are much more tightly coupled to the function than a comment. When I type `help(pvanalytics.quality.irradiance.xyz_qcrad)` I will see the copyright attribution whereas a comment will be stripped out of the code when python loads it.

cwhanse:
wfvining pointed out in an email that the `LICENSES` pattern is used by [pandas](https://github.com/pandas-dev/pandas/tree/master/LICENSES)

wholmgren:
Thanks for the info

>  I prefer docstrings because they are much more tightly coupled to the function than a comment. When I type help(pvanalytics.quality.irradiance.xyz_qcrad) I will see the copyright attribution 

This is what my comment was about. 

> whereas a comment will be stripped out of the code when python loads it.

when `help` loads it yes, but things like pdb and ipython's `??` will show the full source.

> wfvining pointed out in an email that the LICENSES pattern is used by pandas

Seems like a reasonable pattern for the files, but I'm not aware of any copyright notices in the pandas docstrings.

I'm not objecting to what you've done here - just trying to understand. Personally I would not want to follow this pattern in pvlib or SFA unless a project like pandas, numpy, or xarray adopted it.

wfvining:
> wfvining pointed out in an email that the LICENSES pattern is used by pandas

I was referring to the pandas LICENSES directory structure, not the docstrings. I'm not sure how they attribute code taken from other projects.

> when help loads it yes, but things like pdb and ipython's ?? will show the full source.

This makes sense for a source distribution, but is it true of a binary distribution as well? I might be thinking in a `C` mindset where once you "build" the project comments are gone, so in order to satisfy the attribution requirements of an OSS license you have to jump through hoops to ensure that the attribution is retained _even in a binary distribution_.

> Personally I would not want to follow this pattern in pvlib or SFA unless a project like pandas, numpy, or xarray adopted it.

I think this pattern only makes sense for pvanalytics because there are numerous projects we are hoping to incorporate and when we port their code we are substantially reorganizing it (as you can see for what has been ported from SFA).

PR #35, opened 2020-04-21
Trim leading and trailing periods of missing data

wfvining:
Many data-sets come with leading and trailing periods that have sporadic or completely missing data. This adds two functions for identifying and removing these periods:
* `quality.gaps.valid_between()` gives the first and last valid dates in the series.
* `quality.gaps.trim()` returns a boolean mask that is True for the valid period between the dates given by `valid_between`.

These functions both operate on days in the data set, marking all data for each day as valid/invalid based on the number of hours of data that exists on each day. Any data point with a value other than `NaN` is considered "valid" data.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -61,7 +61,7 @@ The following functions identify days with incomplete data.
    quality.gaps.daily_completeness
    quality.gaps.complete
 
-Many data sets may have leading and trailing periods with sparodic or
+Many data sets may have leading and trailing periods of days with sporadic or
 no data. The following functions can be used to remove those periods.
 
 .. autosummary::
@@ -133,4 +133,3 @@ Clearsky
 .. [1]  C. N. Long and Y. Shi, An Automated Quality Assessment and Control
         Algorithm for Surface Radiation Measurements, The Open Atmospheric
         Science Journal 2, pp. 23-37, 2008.
-
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -53,13 +53,21 @@ Identify gaps in the data.
    quality.gaps.interpolation_diff
    quality.gaps.stale_values_diff
 
+The following functions identify days with incomplete data.
+
+.. autosummary::
+   :toctree: generated/
+
+   quality.gaps.daily_completeness
+   quality.gaps.complete
+
 Many data sets may have leading and trailing periods with sparodic or
 no data. The following functions can be used to remove those periods.
 
 .. autosummary::
    :toctree: generated/
 
-   quality.gaps.valid_between
+   quality.gaps.start_stop_dates
    quality.gaps.trim
 
 Outliers
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -58,7 +58,7 @@ The following functions identify days with incomplete data.
 .. autosummary::
    :toctree: generated/
 
-   quality.gaps.daily_completeness
+   quality.gaps.completeness_score
    quality.gaps.complete
 
 Many data sets may have leading and trailing periods of days with sporadic or
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -53,6 +53,15 @@ Identify gaps in the data.
    quality.gaps.interpolation_diff
    quality.gaps.stale_values_diff
 
+Many data sets may have leading and trailing periods with sparodic or
+no data. The following functions can be used to remove those periods.
+
+.. autosummary::
+   :toctree: generated/
+
+   quality.gaps.valid_between
+   quality.gaps.trim
+
 Outliers
 --------
 
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -69,6 +69,7 @@ no data. The following functions can be used to remove those periods.
 
    quality.gaps.start_stop_dates
    quality.gaps.trim
+   quality.gaps.trim_incomplete
 
 Outliers
 --------
--- a/pvanalytics/quality/gaps.py
+++ b/pvanalytics/quality/gaps.py
@@ -277,6 +277,28 @@ def start_stop_dates(series, days=10):
     return start, end
 
 
+def trim(series, days=10):
+    """Mask the begining and end of the data if there are gaps.
+
+    Parameters
+    ----------
+    series : Series
+        A DatetimeIndexed series of booleans
+    days : int, default 10
+        Minimum number of consecutive days that are all True for
+        'good' data to start.
+
+    Returns
+    -------
+    Series
+        A series of booleans with True for all data points between the
+        first and last block of `days` consecutive days that are all
+        True in `series`
+
+    """
+    pass
+
+
 def trim_incomplete(series, minimum_completeness=0.333333, days=10, freq=None):
     """Mask out missing data from the beginning and end of the data.
 
--- a/pvanalytics/quality/gaps.py
+++ b/pvanalytics/quality/gaps.py
@@ -226,33 +226,31 @@ def complete(series, threshold=0.333, freq=None):
     return (completeness >= threshold).reindex(series.index, method='pad')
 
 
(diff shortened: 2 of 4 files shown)


pvanalytics/quality/gaps.py
```diff
@@ -135,3 +136,106 @@ def interpolation_diff(x, window=3, rtol=1e-5, atol=1e-8):
...
+
+    The start and end dates returned by this function can be used to
+    remove large periods of missing data from the begining and end of
+    the series. The valid data begins when there are `days`
+    consecutive days with valid data covering at least `minimum_hours`
+    on each day. Valid data ends on the last day with `days`
+    consecutive days with data covering at least `minimum_hours`
+    preceeding it.
+
+    Any data point with a value other than `NaN` is considered valid
```
cwhanse, line 152: Since we are only removing `NaN`, "valid" is a bit strong.

I would use "valid" for data that also passes physical limit and consistency checks, for example. I also wouldn't use "complete" or "serially complete" since I would expect the output to not have _any_ missing values. Maybe "sufficient"? 

If `series` is 30 days, and there are lots of missing data on day 1, day 15, and day 30, what is returned?  Day 2, Day 14?  Or something else?

wfvining: I think the important thing to describe is what it means for a value to be missing rather than what it means for it to be 'valid.' You are right, "valid" is not the best word to use here.

> If series is 30 days, and there are lots of missing data on day 1, day 15, and day 30, what is returned? Day 2, Day 14? Or something else?

With the default parameters the returned mask will select day 2-29. If there is missing data in the middle of a series nothing happens to it.

wfvining:
Per discussion with Matt and Cliff should expand this to include a function that gives a completeness score to each day (hours per day with data) and a function that gives a boolean mask for days with completeness > some threshold. This mask can/should be used in `trim()`.

wfvining:
Added daily completeness filters and refactored to use these functions in `valid_between`. Renamed `valid_between` to `start_stop_dates` and rewrote documentation to describe the function more accurately.

Tasks
- [ ] add documentation for new functions to API docs
- [ ] fix existing docs (did not change name of `valid_between` there.)

pvanalytics/quality/gaps.py
```diff
@@ -135,3 +136,206 @@ def interpolation_diff(x, window=3, rtol=1e-5, atol=1e-8):
...
+    :py:func:`daily_completeness`
+
+    """
+    complete_days = daily_completeness(series, freq) >= minimum_completeness
+    return complete_days.reindex(series.index, method='pad')
+
+
+def start_stop_dates(series, days=10, minimum_completeness=0.333333,
+                     freq=None):
+    """Get the start and end of data excluding leading and trailing gaps.
```
cwhanse, line 234: Perhaps this function could be made more general by accepting a time series of Boolean, and returning start and stop where the minimum daily count of True is less than the threshold.  That way the criteria for start and stop could change, e.g., fraction of times with irradiance like clear sky.

cwhanse: If we want to keep this function as is, for PVFleets, we could make a `trim_pvfleets` function that implements their workflow.

pvanalytics/quality/gaps.py
```diff
@@ -135,3 +136,206 @@ def interpolation_diff(x, window=3, rtol=1e-5, atol=1e-8):
...
+    ------
+    ValueError
+        If `freq` is longer than the frequency inferred from `series`.
+
+    """
+    inferred_seconds = _freq_to_seconds(pd.infer_freq(series.index))
+    freq_seconds = _freq_to_seconds(freq)
+    seconds_per_sample = freq_seconds or inferred_seconds
+    if freq and inferred_seconds < freq_seconds:
+        raise ValueError("freq must be less than or equal to the"
```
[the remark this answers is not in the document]
wfvining: The rationale behind passing `freq` is that the original time series may have been heavily reduced by previous quality control tests, resulting in an inferred frequency that is much longer than the actual frequency of the data. In that case we would attribute too much time to each data point and calculate a higher completeness than we actually have.

pvanalytics/quality/gaps.py
```diff
@@ -135,3 +136,206 @@ def interpolation_diff(x, window=3, rtol=1e-5, atol=1e-8):
...
+
+    Raises
+    ------
+    ValueError
+        If `freq` is longer than the frequency inferred from `series`.
+
+    """
+    inferred_seconds = _freq_to_seconds(pd.infer_freq(series.index))
+    freq_seconds = _freq_to_seconds(freq)
+    seconds_per_sample = freq_seconds or inferred_seconds
```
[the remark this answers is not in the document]
wfvining: The `or` operator short-circuits, returning `freq_seconds` in that case. I think it isn't clear because as you read the code you assume `_freq_to_seconds` always returns a value (if its argument is `None` then it returns `None`). I will re-work it for clarity.

pvanalytics/quality/gaps.py
wfvining: I like the idea of making this more general is good. If you think passing a boolean series in would achieve that, then I will go ahead and make that change. `trim_pvfleets` (or whatever we want to call it) can easily be implemented using the more general function.

pvanalytics/quality/gaps.py
```diff
@@ -135,3 +136,206 @@ def interpolation_diff(x, window=3, rtol=1e-5, atol=1e-8):
...
+    ValueError
+        See :py:func:`daily_completeness`.
+
+    See Also
+    --------
+    :py:func:`daily_completeness`
+
+    """
+    complete_days = daily_completeness(series, freq) >= minimum_completeness
+    return complete_days.reindex(series.index, method='pad')
```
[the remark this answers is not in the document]
wfvining: I will address this with a kwarg to `daily_completeness` (and rename the function to `completeness_score`).

pvanalytics/quality/gaps.py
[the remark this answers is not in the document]
cwhanse: Or put some inline comments for future maintainers

pvanalytics/quality/gaps.py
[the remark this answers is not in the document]
cwhanse: OK, let's roll with the separate `freq` argument and see how this gets used.

pvanalytics/quality/gaps.py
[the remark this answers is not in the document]
wfvining: I ended up rewriting it. I think there is just a little too much going on for the `arg or default` pattern to be clear.

wfvining:
Added a generic trim function that operates on a series of booleans.

pvanalytics/quality/gaps.py
```diff
@@ -135,3 +136,209 @@ def interpolation_diff(x, window=3, rtol=1e-5, atol=1e-8):
...
+    mask = pd.Series(False, index=series.index)
+    if start:
+        mask.loc[start.date():end.date()] = True
+    return mask
+
+
+def trim_incomplete(series, minimum_completeness=0.333333, days=10, freq=None):
+    """Mask out missing data from the beginning and end of the data.
+
+    Removes data preceeding the start date and following the stop date
```
cwhanse, line 309:
```suggestion
    False for times preceeding the start date and following the stop date
```

docs/api.rst
```diff
@@ -53,6 +53,23 @@ Identify gaps in the data.
...
+   quality.gaps.complete
+
+Many data sets may have leading and trailing periods of days with sporadic or
+no data. The following functions can be used to remove those periods.
+
+.. autosummary::
+   :toctree: generated/
+
+   quality.gaps.start_stop_dates
+   quality.gaps.trim
```
wfvining, line 71: Add `trim_incomplete` here

PR #37, opened 2020-04-23
rounding-based stale value detection

wfvining:
Implementation of the stale value detection method from the pvfleets_qa_analysis code.

This is a very similar function to `stale_values_diff`, but it is based on the difference between consecutive data points after rounding rather than putting a lower bound on the differences. It is also more aggressive in what it considers stale, marking the full sequence of repeated values, rather than just a suffix of the sequence as in `stale_values_diff`.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -54,7 +54,7 @@ Identify gaps in the data.
 
 Data sometimes contains sequences of values that are "stale" or
 "stuck." These are contiguous spans of data where the value does not
-change (or changes by only a very small amount). The functions below
+change within the precision given. The functions below
 can be used to detect stale values.
 
 .. note::
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -51,7 +51,23 @@ Identify gaps in the data.
    :toctree: generated/
 
    quality.gaps.interpolation_diff
+
+Data sometimes contains sequences of values that are "stale" or
+"stuck." These are contiguous spans of data where the value does not
+change (or changes by only a very small amount). The functions below
+can be used to detect stale values.
+
+.. note::
+
+   If the data has been altered in some way (i.e. temperature that has
+   been rounded to an integer value) before being passed to these
+   functions you may see unexpectedly large amounts of stale data.
+
+.. autosummary::
+   :toctree: generated/
+
    quality.gaps.stale_values_diff
+   quality.gaps.stale_values_round
 
 The following functions identify days with incomplete data.
 
--- a/pvanalytics/quality/gaps.py
+++ b/pvanalytics/quality/gaps.py
@@ -162,7 +162,7 @@ def stale_values_round(x, decimals=3, window=4):
     """
     rounded_diff = x.round(decimals=decimals).diff()
     endpoints = rounded_diff.rolling(window=window-1).apply(
-        lambda xs: xs[xs == 0].count() == window-1
+        lambda xs: len(xs[xs == 0]) == window-1
     ).fillna(False).astype(bool)
     flags = endpoints
     while window > 0:
--- a/pvanalytics/quality/gaps.py
+++ b/pvanalytics/quality/gaps.py
@@ -128,31 +128,40 @@ def stale_values_diff(x, window=6, rtol=1e-5, atol=1e-8, mark='tail'):
 def stale_values_round(x, decimals=3, window=6, mark='tail'):
     """Identify stale values by rounding.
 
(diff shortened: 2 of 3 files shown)


wholmgren:
Do you have a real world data set that benefits from this? In practice we find the existing stale filter to be too aggressive to be useful.

wfvining:
I'll see if Matt Muller (who devised this filter) has some specific examples.

pvanalytics/quality/gaps.py
```diff
@@ -86,6 +86,52 @@ def stale_values_diff(x, window=3, rtol=1e-5, atol=1e-8):
     return flags
 
 
+def stale_values_round(x, decimals=3, window=4):
+    """Identify stale values by rounding.
+
+    A value is considered stale if it is part of a sequence of
+    `window` or more consecutive values that are identical when
```
cwhanse, line 93:
```suggestion
   of length `window` of values that are identical when
```

pvanalytics/quality/gaps.py
```diff
@@ -86,6 +86,52 @@ def stale_values_diff(x, window=3, rtol=1e-5, atol=1e-8):
     return flags
 
 
+def stale_values_round(x, decimals=3, window=4):
+    """Identify stale values by rounding.
+
+    A value is considered stale if it is part of a sequence of
+   of length `window` of values that are identical when
```
cwhanse, line 93:
```suggestion
    of length `window` of values that are identical when
```

wfvining:
Matt Muller indicated to me that he hasn't see much of a problem with the sort of false-positives that are noted in the SFA issue, at least not for the data sets he has used. Obviously this approach has the same issues with rounded (or otherwise altered) data (in particular Matt called out rounded temperature data as a problem case). I don't see much of a way around that other than making sure the caller uses a window that is large enough to span more than the expected time for the variable to change.

"More aggressive" was a poor choice of words on my part. All I really meant by it was that `stale_values_rounding()` marks every value in the sequence of values as stale while `stale_values_diff()` leaves the first `window-1` values unmarked. This might actually be a useful feature to add as an option for both functions (perhaps a kwarg `include_prefix=True`).

wfvining:
- [ ] Add `label_all` parameter to control whether every value in the window is labelled (see #43)

docs/api.rst
```diff
@@ -51,7 +51,23 @@ Identify gaps in the data.
    :toctree: generated/
 
    quality.gaps.interpolation_diff
+
+Data sometimes contains sequences of values that are "stale" or
+"stuck." These are contiguous spans of data where the value does not
+change (or changes by only a very small amount). The functions below
```
cwhanse, line 57:
```suggestion
change within the precision given. The functions below
```

pvanalytics/quality/gaps.py
```diff
@@ -125,6 +125,57 @@ def stale_values_diff(x, window=6, rtol=1e-5, atol=1e-8, mark='tail'):
...
+    x : Series
+        Data to be processed.
+    decimals : int, default 3
+        Number of decimal places to round to.
+    window : int, default 6
+        Number of consecutive identical values for a data point to be
+        considered stale.
+    mark : str, default 'tail'
+        How much of the window to mark ``True`` when a sequence of
+        stale values is detected. Can be of 'tail', 'end', or 'all'.
```
cwhanse, line 146:
```suggestion
        stale values is detected. Can be one of 'tail', 'end', or 'all'.
```

pvanalytics/quality/gaps.py
```diff
@@ -125,6 +125,57 @@ def stale_values_diff(x, window=6, rtol=1e-5, atol=1e-8, mark='tail'):
...
+
+        - If 'tail' (the default) then every point in the window
+          *except* the first point is marked ``True``.
+        - If 'end' then only the endpoints of the window are marked
```
cwhanse, line 150: only the last point? I'm confused by "endpoints"

wfvining: This is basically the same as the default behavior for `pandas.Series.rolling` where the value is assigned to the rightmost label in the window. Maybe it would be better to say 'If end then the first `window-1` values in a sequence of stale values are marked False, and all subsequent values in the sequence are marked true'.

pvanalytics/quality/gaps.py
wfvining: I should check other docstrings in gaps, they are probably missing this word as well.

PR #39, opened 2020-04-28
Port derivative-/quantile-based clipping detection from pvfleets QA

wfvining:
Detect clipping in power data by identifying a clipping threshold (through the derivative of the 99.5% quantile of the daily power curve).

All data greater than or equal to the clipping threshold is marked as "clipped".

### DIFF
--- a/pvanalytics/features/clipping.py
+++ b/pvanalytics/features/clipping.py
@@ -102,7 +102,7 @@ def _daytime_powercurve(ac_power):
     power['positive_power'] = ac_power
     power.loc[power.positive_power >= 0, 'positive_power'] = 1
     powerfreq = power.groupby('minutes').positive_power.sum()
-    daytimes = powerfreq[powerfreq > powerfreq.quantile(0.25)].index
+    daytimes = powerfreq[powerfreq >= powerfreq.quantile(0.25)].index
     daytime_power = power[power.minutes.isin(daytimes)]
 
     return daytime_power.groupby('minutes')[power_column].quantile(0.995)
--- a/pvanalytics/features/clipping.py
+++ b/pvanalytics/features/clipping.py
@@ -108,6 +108,12 @@ def _daytime_powercurve(ac_power):
     return daytime_power.groupby('minutes')[power_column].quantile(0.995)
 
 
+def _clipped(power, derivative, clip_derivative, minimum):
+    # test whether a `power` is greater than `minimum` and
+    # `derivative` is less than `clip_derivative`
+    return (np.abs(derivative) <= clip_derivative) and (power > minimum)
+
+
 def _clipping_power(ac_power, clip_derivative=0.0035, freq=None):
     # Returns the clipping power threshold or None if no clipping is
     # identified in the data.
@@ -131,7 +137,7 @@ def _clipping_power(ac_power, clip_derivative=0.0035, freq=None):
     oldpowersum = 0.0
     newpowersum = 0.0
     for derivative, power in zip(power_derivative, powercurve):
-        if ((np.abs(derivative) <= clip_derivative) and (power > power_median * 0.75)):
+        if _clipped(power, derivative, clip_derivative, power_median * 0.75):
             newcount += 1
             newpowersum += power
         else:
--- a/pvanalytics/features/clipping.py
+++ b/pvanalytics/features/clipping.py
@@ -184,4 +184,4 @@ def threshold(ac_power, clip_derivative=0.0035, freq=None):
         clip_derivative=clip_derivative,
         freq=freq
     )
-    return ac_power > threshold
+    return ac_power >= threshold
--- a/pvanalytics/features/clipping.py
+++ b/pvanalytics/features/clipping.py
@@ -174,19 +174,26 @@ def threshold(ac_power, slope_max=0.0035, power_min=0.75,
               freq=None):
     """Detect clipping based on a maximum power threshold.
 
-    A clipping threshold is calculated from the AC power data and any
-    power greater than the threshold is flagged as clipped.
+    This is a two-step process. First a clipping threshold is
+    identified, then any values in `ac_power` greater than or equal
+    to that threshold are flagged.
 
-    To calculate the clipping threshold, `ac_power` is aggregated at
-    each minute of the day. Low power data is removed to eliminate
-    night-time periods and the 99.5% quantile is computed at each
-    minute. If the absolute value of the slope of the 99.5%
-    quantile is less than `slope_max` for
-    a continuous period of at least one hour then clipping is
-    indicated. The mean power for that period is used as the
-    threshold. If there are multiple periods with a slope less
-    than `slope_max` then the longest period is used to compute
-    the threshold.
+    The clipping threshold is determined by computing a 'daily power
+    curve' which is the `power_quantile` quantile of all values in
+    `ac_power` at each minute of the day (excluding night, early morning,
+    and late evening). This gives a rough estimate of the maximum power
+    produced at each minute of the day.
+
+    The minutes of the day where the slope of the daily power curve,
+    normalized by its maximum, is less than `slope_max` are identified. If
+    there is a continuous period of time spanning at least one hour where
+    the slope is less than `slope_max` and the value of the normalized
+    daily power curve is greater than `power_min` times the median of the
+    normalized daily power curve then the data has clipping in it. The
+    average for those minutes in the (not normalized) daily power curve is
+    used as the clipping threshold. If no sufficiently long period with
+    both a low slope and high power exists then there is no clipping in
+    the data.
 
     Parameters
     ----------
@@ -199,17 +206,18 @@ def threshold(ac_power, slope_max=0.0035, power_min=0.75,
         systems.
     power_min: float, default 0.75
         The power during periods with slope less than `slope_max` must
-        be greater than `power_min` time the median daytime power.
+        be greater than `power_min` time the median normalized daytime
+        power.
     power_quantile : float, default 0.995
-        quantile of AC power to consider for determining clipping.
+        Quantile used to calculate the daily power curve.
     frequency_quantile : float, default 0.25
         After taking the count of positive values in `ac_power` at
         each minute of the day, any minute with a count less than the
-        `frequency_quantile`-quantile of all counts is excluded from
-        the calculation of the clipping threshold.
+        `frequency_quantile`-quantile of all counts is excluded. This
+        eliminates early morning/evening times.
     freq : string, default None
         A pandas string offset giving the frequency of data in
-        `ac_power`. If None then the frequency is infered from the
+        `ac_power`. If None then the frequency is inferred from the
         series index.
 
     Returns
--- a/pvanalytics/features/clipping.py
+++ b/pvanalytics/features/clipping.py
@@ -113,11 +113,30 @@ def _clipped(power, derivative, power_min, derivative_max):
     return (np.abs(derivative) <= derivative_max) and (power > power_min)
 
 
-def _clipping_power(ac_power, derivative_max=0.0035,
+def _clipping_power(ac_power, derivative_max=0.0035, power_min=0.75,
                     power_quantile=0.995, frequency_quantile=0.25,
                     freq=None):
-    # Returns the clipping power threshold or None if no clipping is
-    # identified in the data.
+    # Calculate a power threshold, above which the power is being
+    # clipped.
+    #
+    # - The daytime power curve is calculated using
+    #   _daytime_powercurve(). This function groups `ac_power` by
+    #   minute of the day and returns the `power_quantile`-quantile of
+    #   power for each minute, excluding night time, early morning,
+    #   and late afternoon/evening (the amount of data that is
+    #   excluded is controlled by `frequency_quantile`).
+    #
+    # - Each timestamp in the daytime power curve that satisfies the
+    #   clipping criteria[*] is flagged.
+    #
+    # - The clipping threshold is calculated as the mean of the
+    #   longest flagged period in the daytime power curve
+    #
+    # [*] clipping criteria: a timestamp satisfies the clipping
+    # criteria if the derivative of the daytime power curve at that
+    # time is less than `derivative_max` and the value of the daytime
+    # power curve is greater than `power_min` times the median of the
+    # daytime power curve.
     #
     # Copyright (c) 2020 Alliance for Sustainable Energy, LLC.
     if not freq:
@@ -138,7 +157,8 @@ def _clipping_power(ac_power, derivative_max=0.0035,
     longest_powersum = 0
     longest_count = 0
     for derivative, power in zip(power_derivative, powercurve):
-        if _clipped(power, derivative, power_median * 0.75, derivative_max):
+        if _clipped(power, derivative, power_median * power_min,
+                    derivative_max):
             count += 1
             powersum += power
         else:
--- a/pvanalytics/features/clipping.py
+++ b/pvanalytics/features/clipping.py
@@ -175,24 +175,25 @@ def threshold(ac_power, slope_max=0.0035, power_min=0.75,
     """Detect clipping based on a maximum power threshold.
 
     This is a two-step process. First a clipping threshold is
-    identified, then any values in `ac_power` greater than or equal
-    to that threshold are flagged.
+    identified, then any values in `ac_power` greater than or equal to
+    that threshold are flagged.
 
     The clipping threshold is determined by computing a 'daily power
     curve' which is the `power_quantile` quantile of all values in
-    `ac_power` at each minute of the day (excluding night, early morning,
-    and late evening). This gives a rough estimate of the maximum power
-    produced at each minute of the day.
+    `ac_power` at each minute of the day (excluding night, early
+    morning, and late evening). This gives a rough estimate of the
+    maximum power produced at each minute of the day.
 
-    The daily power curve is normalized by its maximum and the minutes of the
-    day are identified where the normalized curve's slope is less than `slope_max`. If
-    there is a continuous period of time spanning at least one hour where
-    the slope is less than `slope_max` and the value of the normalized
-    daily power curve is greater than `power_min` times the median of the
-    normalized daily power curve then the data has clipping in it. If no
-    sufficiently long period with both a low slope and high power exists
-    then there is no clipping in the data.  The average of the daily power
-    curve (not normalized) is the clipping threshold. 
+    The daily power curve is normalized by its maximum and the minutes
+    of the day are identified where the normalized curve's slope is
+    less than `slope_max`. If there is a continuous period of time
+    spanning at least one hour where the slope is less than
+    `slope_max` and the value of the normalized daily power curve is
+    greater than `power_min` times the median of the normalized daily
+    power curve then the data has clipping in it. If no sufficiently
+    long period with both a low slope and high power exists then there
+    is no clipping in the data.  The average of the daily power curve
+    (not normalized) is the clipping threshold.
 
     Parameters
     ----------
--- a/pvanalytics/features/clipping.py
+++ b/pvanalytics/features/clipping.py
@@ -106,14 +106,14 @@ def _daytime_powercurve(ac_power, power_quantile=0.995,
     ).quantile(power_quantile)
 
 
-def _clipped(power, derivative, power_min, derivative_max):
+def _clipped(power, slope, power_min, slope_max):
     # return a mask that is true where `power` is greater than
-    # `power_min` and the absolute value of `derivative` is less
-    # than or equal to `derivative_max`
-    return (np.abs(derivative) <= derivative_max) & (power > power_min)
+    # `power_min` and the absolute value of `slope` is less
+    # than or equal to `slope_max`
+    return (np.abs(slope) <= slope_max) & (power > power_min)
 
 
-def _clipping_power(ac_power, derivative_max=0.0035, power_min=0.75,
+def _clipping_power(ac_power, slope_max=0.0035, power_min=0.75,
                     power_quantile=0.995, frequency_quantile=0.25,
                     freq=None):
     # Calculate a power threshold, above which the power is being
@@ -134,7 +134,7 @@ def _clipping_power(ac_power, derivative_max=0.0035, power_min=0.75,
     #
     # [*] clipping criteria: a timestamp satisfies the clipping
     # criteria if the absolute value of the slope of the daytime power curve
-    # is less than `derivative_max` and the value of the daytime
+    # is less than `slope_max` and the value of the daytime
     # power curve is greater than `power_min` times the median of the
     # daytime power curve.
     #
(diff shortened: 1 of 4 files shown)


wfvining:
Still need to add a tests for:

- [ ] clipping that stops for a brief period in the middle of the day before starting again.
- [ ] multiple days of data (not all of which have clipping)

pvanalytics/features/clipping.py
```diff
@@ -88,3 +88,115 @@ def levels(ac_power, window=4, fraction_in_window=0.75,
...
+    normalized_power = powercurve / powercurve.max()
+    power_derivative = (normalized_power.diff()
+                        / normalized_power.index.to_series().diff()) * freq
+    power_median = powercurve.median()
+
+    powersum = 0
+    count = 0
+    longest_powersum = 0
+    longest_count = 0
+    for derivative, power in zip(power_derivative, powercurve):
```
cwhanse, line 139: seems like a use for `Series.rolling` somehow.

pvanalytics/features/clipping.py
```diff
@@ -88,3 +88,115 @@ def levels(ac_power, window=4, fraction_in_window=0.75,
...
+        if _clipped(power, derivative, clip_derivative, power_median * 0.75):
```
cwhanse, line 140: the 0.75 should be an input.

pvanalytics/features/clipping.py
```diff
@@ -88,3 +88,115 @@ def levels(ac_power, window=4, fraction_in_window=0.75,
...
+def _clipped(power, derivative, clip_derivative, minimum):
+    # test whether a `power` is greater than `minimum` and
+    # `derivative` is less than `clip_derivative`
+    return (np.abs(derivative) <= clip_derivative) and (power > minimum)
+
+
+def _clipping_power(ac_power, clip_derivative=0.0035, freq=None):
+    # Returns the clipping power threshold or None if no clipping is
+    # identified in the data.
+    #
```
cwhanse, line 120: This function needs a description of the algorithm it implements.

pvanalytics/features/clipping.py
```diff
@@ -88,3 +88,115 @@ def levels(ac_power, window=4, fraction_in_window=0.75,
...
+def threshold(ac_power, clip_derivative=0.0035, freq=None):
+    """Detect clipping based on a maximum power threshold.
+
+    A clipping threshold is calculated from the AC power data and any
+    power greater than the threshold is flagged as clipped.
+
+    To calculate the clipping threshold, `ac_power` is aggregated at
+    each minute of the day. Low power data is removed to eliminate
+    night-time periods and the 99.5% quantile is computed. If the
+    derivative of the 99.5% quantile is less than `clip_derivative`
```
cwhanse, line 166:
```suggestion
    absolute value of the derivative of the 99.5% quantile (with respect to time) is less than `clip_derivative`
```

pvanalytics/features/clipping.py
```diff
@@ -88,3 +88,115 @@ def levels(ac_power, window=4, fraction_in_window=0.75,
...
+    power['positive_power'] = ac_power
+    power.loc[power.positive_power >= 0, 'positive_power'] = 1
+    powerfreq = power.groupby('minutes').positive_power.sum()
+    daytimes = powerfreq[powerfreq >= powerfreq.quantile(0.25)].index
+    daytime_power = power[power.minutes.isin(daytimes)]
+
+    return daytime_power.groupby('minutes')[power_column].quantile(0.995)
+
+
+def _clipped(power, derivative, clip_derivative, minimum):
```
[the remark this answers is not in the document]
cwhanse: Same renaming throughout if you agree

pvanalytics/features/clipping.py
```diff
@@ -88,3 +88,115 @@ def levels(ac_power, window=4, fraction_in_window=0.75,
...
+
+def threshold(ac_power, clip_derivative=0.0035, freq=None):
+    """Detect clipping based on a maximum power threshold.
+
+    A clipping threshold is calculated from the AC power data and any
+    power greater than the threshold is flagged as clipped.
+
+    To calculate the clipping threshold, `ac_power` is aggregated at
+    each minute of the day. Low power data is removed to eliminate
+    night-time periods and the 99.5% quantile is computed. If the
```
cwhanse, line 165:
```suggestion
    night-time periods and the 99.5% quantile is computed at each minute. If the
```

pvanalytics/features/clipping.py
```diff
@@ -88,3 +88,115 @@ def levels(ac_power, window=4, fraction_in_window=0.75,
...
+    threshold. If there are multiple periods with a derivative less
+    than `clip_derivative` then the longest period is used to compute
+    the threshold.
+
+    Parameters
+    ----------
+    ac_power : Series
+        DatetimeIndexed series of AC power data.
+    clip_derivative : float, default 0.0035
+        Minimum derivative for clipping to be indicated. The default
```
cwhanse, line 178:
```suggestion
        Maximum absolute value of derivative of AC power quantile for clipping to be indicated. The default
```

pvanalytics/features/clipping.py
```diff
@@ -88,3 +88,115 @@ def levels(ac_power, window=4, fraction_in_window=0.75,
...
+
+    Returns
+    -------
+    Series
+        True when clipping is indicated.
+
+    Notes
+    -----
+    This function is based on code from the pvfleets_qa_analysis
+    project. Copyright (c) 2020 Alliance for Sustainable Energy, LLC.
```
cwhanse, line 194: Here, I'm now unsure about the copyright. The algorithm is from pvfleets but the code is original here.

pvanalytics/features/clipping.py
wfvining: I couldn't see a way to do it with Series.rolling because the count resets under certain conditions. It is kind of a weird combination of `Series.expanding` and `Series.rolling` I will dig in a little deeper to both and see if there is a way to get rid of the loop.

pvanalytics/features/clipping.py
```diff
@@ -88,3 +88,115 @@ def levels(ac_power, window=4, fraction_in_window=0.75,
...
+    power_column = power.columns[0]
+    power['minutes'] = power.index.hour * 60 + power.index.minute
+
+    power['positive_power'] = ac_power
+    power.loc[power.positive_power >= 0, 'positive_power'] = 1
+    powerfreq = power.groupby('minutes').positive_power.sum()
+    daytimes = powerfreq[powerfreq >= powerfreq.quantile(0.25)].index
+    daytime_power = power[power.minutes.isin(daytimes)]
+
+    return daytime_power.groupby('minutes')[power_column].quantile(0.995)
```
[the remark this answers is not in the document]
wfvining: It is not necessary. This is how it was done in the pvfleets code, I've learned enough pandas now that I can rewrite it without converting to a DataFrame.

pvanalytics/features/clipping.py
```diff
@@ -88,3 +88,115 @@ def levels(ac_power, window=4, fraction_in_window=0.75,
...
+    return (np.abs(derivative) <= clip_derivative) and (power > minimum)
+
+
+def _clipping_power(ac_power, clip_derivative=0.0035, freq=None):
+    # Returns the clipping power threshold or None if no clipping is
+    # identified in the data.
+    #
+    # Copyright (c) 2020 Alliance for Sustainable Energy, LLC.
+    if not freq:
+        freq = pd.Timedelta(pd.infer_freq(ac_power.index)).seconds * 60
```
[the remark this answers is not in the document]
wfvining: A few lines later the units are Power/minute, we multiply by `freq` (in minutes) to put the data back in units of Power.

pvanalytics/features/clipping.py
wfvining: I have a nice solution to this that uses cumsum and some other tricks to make the code more readable. Probably doesn't have much of a performance impact since the loop was executing at most 1440 iterations (# of minutes in the day).

pvanalytics/features/clipping.py
wfvining: I'm also unsure. The code here continues to diverge, so maybe we remove the copyright notice and just credit that it is based on pvfleest_qa_analysis?

cwhanse: Yes, where the method is the same but the implementation is substantially different. Where the pvfleets code is reused (with minor changes such as variable names) we should carry their copyright.

pvanalytics/features/clipping.py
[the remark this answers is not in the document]
cwhanse: To get minutes would be " / 60" here?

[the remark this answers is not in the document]
wfvining: \*doh\*, thanks.

pvanalytics/features/clipping.py
```diff
@@ -88,3 +88,141 @@ def levels(ac_power, window=4, fraction_in_window=0.75,
...
+                    freq=None):
+    # Calculate a power threshold, above which the power is being
+    # clipped.
+    #
+    # - The daytime power curve is calculated using
+    #   _daytime_powercurve(). This function groups `ac_power` by
+    #   minute of the day and returns the `power_quantile`-quantile of
+    #   power for each minute, excluding night time, early morning,
+    #   and late afternoon/evening (the amount of data that is
+    #   excluded is controlled by `frequency_quantile`).
```
cwhanse, line 127: use `:py:func:` < full path to frequency_quantile > `

pvanalytics/features/clipping.py
```diff
@@ -88,3 +88,141 @@ def levels(ac_power, window=4, fraction_in_window=0.75,
...
+    #
+    # - Each timestamp in the daytime power curve that satisfies the
+    #   clipping criteria[*] is flagged.
+    #
+    # - The clipping threshold is calculated as the mean of the
```
cwhanse, line 132:
```suggestion
    # - The clipping threshold is calculated as the mean power during the
```

pvanalytics/features/clipping.py
```diff
@@ -88,3 +88,141 @@ def levels(ac_power, window=4, fraction_in_window=0.75,
...
+    #   longest flagged period in the daytime power curve
```
cwhanse, line 133:
```suggestion
    #   longest flagged period in the daytime power curve.
```

pvanalytics/features/clipping.py
```diff
@@ -88,3 +88,141 @@ def levels(ac_power, window=4, fraction_in_window=0.75,
...
+    #
+    # [*] clipping criteria: a timestamp satisfies the clipping
+    # criteria if the derivative of the daytime power curve at that
+    # time is less than `derivative_max` and the value of the daytime
+    # power curve is greater than `power_min` times the median of the
+    # daytime power curve.
```
cwhanse, line 139:
```suggestion
    # criteria if the absolute value of the slope of the daytime power curve
    # is less than `derivative_max` and the value of the daytime
    # power curve is greater than `power_min` times the median of the
    # daytime power curve.
```

pvanalytics/features/clipping.py
```diff
@@ -88,3 +88,141 @@ def levels(ac_power, window=4, fraction_in_window=0.75,
...
+
+    # Use the derivative of the 99.5% quantile of daytime power at
+    # each minute to identify clipping.
+    powercurve = _daytime_powercurve(ac_power)
+    normalized_power = powercurve / powercurve.max()
+    power_derivative = (normalized_power.diff()
+                        / normalized_power.index.to_series().diff()) * freq
+
+    clipped_times = _clipped(powercurve, power_derivative,
+                             powercurve.median() * power_min,
```
cwhanse, line 155: I think the intent was to apply the median at each minute (of the day), is `powercurve.median()` is the median of the `powercurve` time series?

pvanalytics/features/clipping.py
```diff
@@ -88,3 +88,141 @@ def levels(ac_power, window=4, fraction_in_window=0.75,
...
+                             derivative_max)
+    clipping_cumsum = (~clipped_times).cumsum()
+    # get the value of the cumulative sum the longest True span
```
cwhanse, line 158:
```suggestion
    # get the value of the cumulative sum of the longest True span
```

pvanalytics/features/clipping.py
```diff
@@ -88,3 +88,141 @@ def levels(ac_power, window=4, fraction_in_window=0.75,
...
+              freq=None):
+    """Detect clipping based on a maximum power threshold.
+
+    A clipping threshold is calculated from the AC power data and any
+    power greater than the threshold is flagged as clipped.
+
+    To calculate the clipping threshold, `ac_power` is aggregated at
+    each minute of the day. Low power data is removed to eliminate
+    night-time periods and the 99.5% quantile is computed at each
+    minute. If the absolute value of the derivative of the 99.5%
```
cwhanse, line 181:
```suggestion
    minute. If the absolute value of the slope of the 99.5%
```

pvanalytics/features/clipping.py
```diff
@@ -88,3 +88,141 @@ def levels(ac_power, window=4, fraction_in_window=0.75,
...
+    quantile (with respect to time) is less than `derivative_max` for
```
cwhanse, line 182:
```suggestion
    quantile is less than `derivative_max` for
```

pvanalytics/features/clipping.py
```diff
@@ -88,3 +88,141 @@ def levels(ac_power, window=4, fraction_in_window=0.75,
...
+    a continuous period of at least one hour then clipping is
+    indicated. The mean power for that period is used as the
+    threshold. If there are multiple periods with a derivative less
+    than `derivative_max` then the longest period is used to compute
```
cwhanse, line 186:
```suggestion
    threshold. If there are multiple periods with a slope less
    than `derivative_max` then the longest period is used to compute
```

pvanalytics/features/clipping.py
```diff
@@ -88,3 +88,141 @@ def levels(ac_power, window=4, fraction_in_window=0.75,
...
+    longest = powercurve[clipping_cumsum == longest_clipped]
+
+    if longest.index[-1] - longest.index[0] >= 60:
+        # if the period of clipping is at least 60 minutes then we
+        # have enough evidence to determine the clipping threshold.
+        return longest.mean()
+    return None
+
+
+def threshold(ac_power, derivative_max=0.0035,
```
cwhanse, line 170: I'm inclined to ask for `derivative` and `derivative_max` to be `slope` and `slope_max` respectively. Slope is (to me) unambiguous, whereas a numerical derivative can become quite complicated in its calculation.

pvanalytics/features/clipping.py
```diff
@@ -88,3 +88,141 @@ def levels(ac_power, window=4, fraction_in_window=0.75,
...
+    indicated. The mean power for that period is used as the
+    threshold. If there are multiple periods with a derivative less
+    than `derivative_max` then the longest period is used to compute
+    the threshold.
+
+    Parameters
+    ----------
+    ac_power : Series
+        DatetimeIndexed series of AC power data.
+    derivative_max : float, default 0.0035
```
cwhanse, line 193: This needs a statement that it's a fraction of AC power normalized by maximum AC power.

pvanalytics/features/clipping.py
```diff
@@ -88,3 +88,141 @@ def levels(ac_power, window=4, fraction_in_window=0.75,
...
+        Maximum absolute value of derivative of AC power quantile for
+        clipping to be indicated. The default value has been derived
+        empirically to prevent false positives for tracking PV
+        systems.
+    power_quantile : float, default 0.995
+        quantile used to determine the maximum power for each minute
+        of the day
```
cwhanse, line 200:
```suggestion
        quantile of AC power to consider for determining clipping.
```

pvanalytics/features/clipping.py
```diff
@@ -88,3 +88,141 @@ def levels(ac_power, window=4, fraction_in_window=0.75,
...
+    frequency_quantile : float, default 0.25
+        After taking the count of positive values in `ac_power` at
+        each minute of the day, any minute with a count less than the
+        `frequency_quantile`-quantile of all counts is excluded from
+        the calculation of the clipping threshold.
```
cwhanse, line 205: Is this a reference to the day/ night detector?

pvanalytics/features/clipping.py
```diff
@@ -88,3 +88,140 @@ def levels(ac_power, window=4, fraction_in_window=0.75,
...
+    clipped_times = _clipped(powercurve, power_derivative,
+                             powercurve.median() * power_min,
+                             derivative_max)
+    clipping_cumsum = (~clipped_times).cumsum()
+    # get the value of the cumulative sum of the longest True span
+    longest_clipped = clipping_cumsum.value_counts().idxmax()
+    # select the longest span that satisfies the clipping criteria
+    longest = powercurve[clipping_cumsum == longest_clipped]
+
+    if longest.index[-1] - longest.index[0] >= 60:
```
wfvining, line 163: Is it better to use `longest.index.min()` and `longest.index.max()` here since there is nothing that guarantees indexes are in order?

pvanalytics/features/clipping.py
wfvining: Sounds good to me.

pvanalytics/features/clipping.py
wfvining: This is a parameter, actually. We apply two quantiles: `power_quantile` which forms the power curve, and `frequency_quantile`.  I made a mistake here and forgot to pass it to `_daytime_powercurve`.

pvanalytics/features/clipping.py
wfvining: I'm glad you brought this up. We do a sort of day/night filtering step in `_daytime_powercurve`. It might be better to pass in only daytime power, or pass a day/night mask. Passing a mask is probably the best solution to limit coupling between modules while still reducing code duplication.

@matt14muller, are there any caveats/pitfalls you can see in changing the original algorithm like this?

pvanalytics/features/clipping.py
wfvining: It is the median for the whole power curve. The intent is to not mark periods of low power as clipping. This codifies the assumption that clipping will always be in periods of relatively high power.

pvanalytics/features/clipping.py
```diff
@@ -88,3 +88,146 @@ def levels(ac_power, window=4, fraction_in_window=0.75,
...
+    ac_power : Series
+        DatetimeIndexed series of AC power data.
+    slope_max : float, default 0.0035
+        Maximum absolute value of slope of AC power quantile for
+        clipping to be indicated. The default value has been derived
+        empirically to prevent false positives for tracking PV
+        systems.
+    power_min: float, default 0.75
+        The power during periods with slope less than `slope_max` must
+        be greater than `power_min` time the median daytime power.
```
wfvining, line 202: The description of this parameter needs to be improved. I'm having a hard time finding the words at the moment, will come back to it later.

pvanalytics/features/clipping.py
wfvining: Fixed this in ba688a5

pvanalytics/features/clipping.py
wfvining: I think it is actually the daily power curve (99.5% quantile of `ac_power` at each minute of the day) that is normalized. I'm not sure how to communicate that in the description of the parameter. I think if I spend a bit of time writing a better _overall_ description of the algorithm (and define names like "daily power curve") it will help with describing the parameters.

wfvining: I have rewritten the description of how the clipping threshold is calculated, hopefully in doing so the meaning of all the parameters is also much more clear, but I will need to re-read it after some time working on something else to know for sure...

pvanalytics/features/clipping.py
wfvining: I'm actually starting to wonder if this step is even necessary. We only accept a period with low slope a clipping if it has sufficiently high power (see the `power_min` parameter). That should prevent using night time or early morning periods where we shouldn't be seeing clipping as the the clipping threshold.

matt14muller: I am not sure I follow your question.   Can you elaborate?

wfvining: There is a step where minutes of the day with a low number of positive values are eliminated. I think it is to remove the night/early morning/evening times. Do you think that is really necessary? Can we just include night-time/morning/evening when looking for clipping?

pvanalytics/features/clipping.py
cwhanse: I'm okay with this

pvanalytics/features/clipping.py
```diff
@@ -88,3 +88,154 @@ def levels(ac_power, window=4, fraction_in_window=0.75,
...
+    to that threshold are flagged.
+
+    The clipping threshold is determined by computing a 'daily power
+    curve' which is the `power_quantile` quantile of all values in
+    `ac_power` at each minute of the day (excluding night, early morning,
+    and late evening). This gives a rough estimate of the maximum power
+    produced at each minute of the day.
+
+    The minutes of the day where the slope of the daily power curve,
+    normalized by its maximum, is less than `slope_max` are identified. If
```
cwhanse, line 188:
```suggestion
    The daily power curve is normalized by its maximum and the minutes of the
    day are identified where the normalized curve's slope is less than `slope_max`. If
```

pvanalytics/features/clipping.py
```diff
@@ -88,3 +88,154 @@ def levels(ac_power, window=4, fraction_in_window=0.75,
...
+    there is a continuous period of time spanning at least one hour where
+    the slope is less than `slope_max` and the value of the normalized
+    daily power curve is greater than `power_min` times the median of the
+    normalized daily power curve then the data has clipping in it. The
+    average for those minutes in the (not normalized) daily power curve is
+    used as the clipping threshold. If no sufficiently long period with
+    both a low slope and high power exists then there is no clipping in
+    the data.
```
cwhanse, line 196:
```suggestion
    normalized daily power curve then the data has clipping in it. If no
    sufficiently long period with both a low slope and high power exists
    then there is no clipping in the data.  The average of the daily power
    curve (not normalized) is the clipping threshold. 
```

pvanalytics/features/clipping.py
cwhanse: >  Can we just include night-time/morning/evening when looking for clipping?

We can with this algorithm, since there is a power threshold critieria. The SFA function only looks for flat spots, and has to be combined with a night/day detector to avoid labeling night as clipping.

wfvining: I'll need to remove the `frequency_quantile` parameter (and associated code) and see if the tests still pass.

wfvining: @cwhanse @matt14muller I removed the `frequency_quantile` param. Works fine for the tests, but those are just simple tests... The only problem I can think of with this is if there is enough erroneous high data in the night time that the 95% quantile is driven up. I'm still good with eliminating the parameter though since we have many other quality functions to help users eliminate that problem. If worst comes to worst, users can simply pass only daytime power values, leaving night time out of the equation all together.

PR #40, opened 2020-05-12
Functions to detect when the sun is up

wfvining:
Two methods to identify when the sun is up from power or irradiance data. These are useful when you do not trust/know the timezone of timestamps in your data. 

Both functions work by aggregating data by minute of the day
* `features.daylight.frequency()` looks for the minutes of the day where the vast majority of days (defaults to 80%) in the data have positive values.
* `features.daylight.level()` looks for the minutes of the day where the mean of all data for that minute is greater than some percentage (defaults to 20%) of some quantile (defaults to 95%) of the data.

### To Do
- [x] Add to API documentation
- [x] move to `features.daylight` submodule
- [x] Consider additional tests (did I miss any corner cases?).
    - Add a test where the daytime period is split across two days in the index
- [x] Update readme to reflect the new modules/submodules


wfvining:
@matt14muller let me know if you have any thoughts on the `daytime_level()` function.

wfvining:
Our original plan (see README.md) was to put these types of functions in the `filtering` module. Should we do that, or do we want to keep them here. Personally, I think it makes more sense to put these in features since they are identifying a feature in the data, but I could be convinced otherwise.

cwhanse:
I'm OK with putting it these functions in `features`, maybe make a `daylight` folder within `features`

matt14muller:
I have been doing some testing of what Kirsten wrote using derivates and integrals to identify day/night/and clipping and so far the results look good.  I am working on some edge case testing to see if they cause problems.
Also Kirsten said she is happy to participate.

Matt

wfvining:
`features.daylight.[frequency|level]` seems good to me. I didn't really like `features.time` now that you mention it.

docs/api.rst
```diff
@@ -129,6 +129,27 @@ Clearsky
 
    features.clearsky.reno
 
+Time
+----
+
+The following functions can be used to differentiate night-time and
+day-time based on power or irradiance data. This is useful if you do
+not know the time-zone of the index for your data of for verifying
```
cwhanse, line 137:
```suggestion
not know the timezone of the index for your data or for verifying
```

docs/api.rst
```diff
@@ -129,6 +129,27 @@ Clearsky
...
+that the time-zone is correct. Both functions identify the same
```
cwhanse, line 138:
```suggestion
that the timezone is correct. Both functions identify the same
```

docs/api.rst
```diff
@@ -129,6 +129,27 @@ Clearsky
...
+feature, but in slightly different ways.
+:py:func:`features.daylight.frequency` is useful if your data has
+substantial outliers or other excessively large values; however, it
+may fail if substantial portions of the night time data is greater
```
cwhanse, line 142:
```suggestion
may fail if substantial portions of the night-time data is greater
```

pvanalytics/features/daylight.py
```diff
@@ -0,0 +1,97 @@
...
+    data[data <= 0] = np.nan
+    minutes = pd.Series(_minute_of_day(data.index), index=data.index)
+    # group by minute of day and compute frequency of positive values
+    # at each minute
+    value_frequency = data.groupby(minutes).count()
+    if value_frequency.max() < minimum_days:
+        raise ValueError("Too few days with data (got {}, minimum_days={})"
+                         .format(value_frequency.max(), minimum_days))
+    daylight_minutes = value_frequency[
+        value_frequency > threshold * value_frequency.mean()
```
cwhanse, line 58: @matt14muller I don't understand how this works. `value_frequency[minute_of_day]` is the count of days when the data value is positive. `value_frequency.mean()` is the average of these counts. What happens if the data span a summer a high latitude? Since the daylight period is long, `value_frequency.mean()` will be large compared to data from a winter at the same latitude. So it is more likely to label a minute as daytime during winter than summer?

pvanalytics/features/daylight.py
```diff
@@ -0,0 +1,97 @@
...
+    return minutes.isin(daylight_minutes)
+
+
+def level(power_or_irradiance, threshold=0.2, quantile=0.95):
+    """Identify daytime periods based on a minimum power/irradiance threshold.
+
+    Power or irradiance data is aggregated by minute of day and times
+    where the mean power is greater than `threshold` * max are marked
+    as daytime, where max is the `quantile`-percent quantile of the
+    data.
```
cwhanse, line 70:
```suggestion
    Power or irradiance data is aggregated by minute of day and the
    mean is computed at each minute. A minute is marked as daytime
    when the mean value is greater than or equal to `threshold` times
    the `quantile`-percent of `power_or_irradiance`.
 ```

File: pvanalytics/features/daylight.py [hunk 6]
```diff
@@ -0,0 +1,97 @@
...
+        Quantile to use as the upper bound of the data.
+
+    Returns
+    -------
+    Series
+        A series of booleans with True for timestamps when the sun is
+        up.
+
+    """
+    max_power = power_or_irradiance.quantile(quantile)
```
cwhanse, line 90: @matt14muller `max_power` is the 95th quantile of the time series, e.g. a value representative of solar noon. So it is labeling daylight as minutes when the time series is greater than or equal to the 95th percentile. The labels from this algorithm are going to depend on the season, e.g., early morning minutes are less likely to be labeled in the summer than in the winter since the solar noon peak is higher in summer.

wfvining: I wrote this one based on our discussion a few weeks ago (not based on the pvfleets code, even though it is structurally similar). I had not thought of the issue you pointed out with the higher values in the summer, that might be a good reason not to include this. If you only really want mid-day periods and you don't mind losing the morning/evening data then I do think that this algorithm works well (even on the wacky irradiance data set that Matt shared with me).

wfvining:
Closing this since #67 implements a better solution.

PR #46, opened 2020-05-26
Optionally label every interpolated/stale point (not just endpoints)

### LINKED ISSUE #43: stale_values_diff and interpolation_diff : option to label all values
Consider adding an option to label all values in a stale / interpolation window as stale, rather than only labelling the end value in a window.

wfvining:
Closes #43 

Adds kwarg `label_all` to `gaps.stale_values_diff` and `gaps.interpolation_diff`. If `label_all=True` then every point in a window that appears interpolated or stale is marked `True`. If `label_all=False` then only the endpoint of the window is marked `True`.

### DIFF
--- a/pvanalytics/quality/gaps.py
+++ b/pvanalytics/quality/gaps.py
@@ -55,6 +55,9 @@ def stale_values_diff(x, window=3, rtol=1e-5, atol=1e-8, label_all=False):
     stale if all values in the window are close to the first value
     (index 0).
 
+    Parameters `rtol` and `atol` have the same meaning as in
+    :py:func:`numpy.allclose`.
+
     Parameters
     ----------
     x : Series
@@ -69,9 +72,6 @@ def stale_values_diff(x, window=3, rtol=1e-5, atol=1e-8, label_all=False):
     label_all : bool, default False
         Whether to label the full window. If False, then just the
         endpoints of the window are labeled.
-
-    Parameters rtol and atol have the same meaning as in
-    numpy.allclose
 
     Returns
     -------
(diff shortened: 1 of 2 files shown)


pvanalytics/quality/gaps.py
```diff
@@ -35,13 +35,29 @@ def _all_close_to_first(x, rtol=1e-5, atol=1e-8):
     return np.allclose(a=x, b=x[0], rtol=rtol, atol=atol)
 
 
-def stale_values_diff(x, window=3, rtol=1e-5, atol=1e-8):
+def _backfill_window(endpoints, window):
+    # propagate Trues in `endpoints` back `window` periods.  This
+    # makes Trues fill the entire window, rather than just marking the
+    # endpoints of the window.
```
cwhanse, line 41:
```suggestion
    # right endpoint of each window.
```

pvanalytics/quality/gaps.py
```diff
@@ -53,9 +69,9 @@ def stale_values_diff(x, window=3, rtol=1e-5, atol=1e-8):
         relative tolerance for detecting a change in data values
     atol : float, default 1e-8
         absolute tolerance for detecting a change in data values
-
-    Parameters rtol and atol have the same meaning as in
-    numpy.allclose
+    label_all : bool, default False
+        Whether to label the full window. If False, then just the
+        endpoints of the window are labeled.
```
cwhanse, line 74:
```suggestion
        Whether to label the full window. If False, then only the right
        endpoint of the window is labeled.
```

cwhanse: If the edit is accepted here, also applies to `interpolation_diff`

pvanalytics/quality/gaps.py
```diff
@@ -105,6 +126,9 @@ def interpolation_diff(x, window=3, rtol=1e-5, atol=1e-8):
         tolerance relative to max(abs(x.diff()) for detecting a change
     atol : float, default 1e-8
         absolute tolerance for detecting a change in first difference
+    label_all : bool, default False
+        Whether to label all values in the window. If False only the
+        endpoints of the window are labeled.
```
wfvining, line 131:
```suggestion
        Whether to label all values in the window. If False, then only the
        right endpoint of the window is labeled.
```

PR #49, opened 2020-05-29
Determine whether a system has a fixed orientation or is equipped with a tracker

wfvining:
This adds a function to the `pvanalytics.system` module for identifying the orientation (Fixed or Tracking) of a PV system from power or irradiance data. This is a fairly early draft, but I wanted to get it out there for early feedback/suggestions.

This is derived from the PVFleets QA Analysis project.

## (incomplete) To Do list
- [x] finish implementing testing plan (see comments in `tests/test_system.py`)
- [x] decide whether/how to pass goodness of fit parameters
- [x] add to API documentation
- [x] test winter and summer months separately. This is necessary in order to catch problems during the winter that are masked when the entire series is considered as one because the maximum power/irradiance envelope is higher in the summer months.

### DIFF
--- a/pvanalytics/system.py
+++ b/pvanalytics/system.py
@@ -25,9 +25,10 @@ def _group_by_minute(data):
 def _fit_quadratic(data):
     # Fit a quadratic to `data` returning R^2 for the fit.
     coefficients = np.polyfit(data.index, data, 2)
+    quadratic = np.poly1d(coefficients)
     # Calculate the R^2 for the fit
     _, _, correlation, _, _ = scipy.stats.linregress(
-        data, np.poly1d(coefficients)
+        data, quadratic(data.index)
     )
     return correlation**2
 
--- a/pvanalytics/system.py
+++ b/pvanalytics/system.py
@@ -1,8 +1,6 @@
 """Functions for identifying system characteristics."""
 import enum
-import numpy as np
-import scipy.stats
-import scipy.optimize
+from pvanalytics.util import _fit, _group
 
 
 @enum.unique
@@ -30,45 +28,6 @@ def _remove_morning_evening(data, threshold):
     # Remove morning and evening data by excluding times where
     # `data` is less than ``threshold * data.max()``
     return data[data > threshold * data.max()]
-
-
-def _group_by_minute(data):
-    # return data grouped by minute of the day.
-    return data.groupby(data.index.hour * 60 + data.index.minute)
-
-
-def _fit_quadratic(data):
-    # Fit a quadratic to `data` returning R^2 for the fit.
-    coefficients = np.polyfit(data.index, data, 2)
-    quadratic = np.poly1d(coefficients)
-    # Calculate the R^2 for the fit
-    _, _, correlation, _, _ = scipy.stats.linregress(
-        data, quadratic(data.index)
-    )
-    return correlation**2
-
-
-def _fit_quartic(data, middle):
-    # Fit a quartic to the data.
-    #
-    # The quartic must
-    # - open downwards
-    # - centered within 70 minutes of the middle of `data`
-    # - the y-value of the middle must be within 15% of the median of `data`
-    #
-    # Returns the R^2 for the fit.
-    def quartic(x, a, b, c, e):
-        return a * (x - e)**4 + b * (x - e)**2 + c
-    median = data.median()
-    params, _ = scipy.optimize.curve_fit(
-        quartic,
-        data.index, data,
-        bounds=((-1e-05, 0, median * 0.85, middle - 70),
-                (-1e-10, median * 3e-05, median * 1.15, middle + 70))
-    )
-    model = quartic(data.index, params[0], params[1], params[2], params[3])
-    residuals = data - model
-    return 1 - (np.sum(residuals**2) / np.sum((data - np.mean(data))**2))
 
 
 def _orientation_from_fit(rsquared_quadratic, rsquared_quartic,
@@ -132,12 +91,12 @@ def orientation(series, daytime, clipping, clip_max=10.0,
     """
     fit_params = fit_params or PVFLEETS_FIT_PARAMS
     envelope = _remove_morning_evening(
-        _group_by_minute(series[daytime]).quantile(0.995),
+        _group.by_minute(series[daytime]).quantile(0.995),
         0.05
     )
     middle = (envelope.index.max() + envelope.index.min()) / 2
-    rsquared_quadratic = _fit_quadratic(envelope)
-    rsquared_quartic = _fit_quartic(envelope, middle)
+    rsquared_quadratic = _fit.quadratic(envelope)
+    rsquared_quartic = _fit.quartic_restricted(envelope, middle)
     system_orientation = _orientation_from_fit(
         rsquared_quadratic, rsquared_quartic,
         (clipping[daytime].sum() / len(clipping[daytime])) * 100,
@@ -146,15 +105,15 @@ def orientation(series, daytime, clipping, clip_max=10.0,
     )
     if fit_median:
         median = _remove_morning_evening(
-            _group_by_minute(series[daytime]).median(),
+            _group.by_minute(series[daytime]).median(),
             0.025
         )
         if system_orientation is Orientation.FIXED:
-            quadratic_median = _fit_quadratic(median)
+            quadratic_median = _fit.quadratic(median)
             if quadratic_median < 0.9:
                 return Orientation.UNKNOWN
         elif system_orientation is Orientation.TRACKING:
-            quartic_median = _fit_quartic(median, middle)
+            quartic_median = _fit.quartic_restricted(median, middle)
             if quartic_median < 0.9:
                 return Orientation.UNKNOWN
     return system_orientation
--- a/pvanalytics/system.py
+++ b/pvanalytics/system.py
@@ -163,8 +163,8 @@ def is_tracking_envelope(series, daytime, clipping, clip_max=0.1,
         clipped.
     clip_max : float, default 0.1
         If the fraction of data flagged as clipped is greater than
-        `clip_max` then tracking cannot be determined and
-        :py:const:`Tracker.UNKNOWN` is returned.
+        `clip_max` then it cannot be determined whether the system is
+        tracking or fixed and :py:const:`Tracker.UNKNOWN` is returned.
     envelope_quantile : float, default 0.995
         Quantile used to determine the upper power or irradiance
         envelope.
--- a/pvanalytics/system.py
+++ b/pvanalytics/system.py
@@ -1,5 +1,8 @@
 """Functions for identifying system characteristics."""
 import enum
+import numpy as np
+import pandas as pd
+import pvlib
 from pvanalytics.util import _fit, _group
 
 
@@ -91,8 +94,8 @@ def _infer_tracking(series, envelope_quantile,
         envelope_min_fraction
     )
     middle = (envelope.index.max() + envelope.index.min()) / 2
-    rsquared_quadratic = _fit.quadratic(x=envelope.index, y=envelope)
-    rsquared_quartic = _fit.quartic_restricted(
+    rsquared_quadratic = _fit.quadratic_r2(x=envelope.index, y=envelope)
+    rsquared_quartic = _fit.quartic_restricted_r2(
         x=envelope.index,
         y=envelope,
         noon=middle
@@ -107,11 +110,11 @@ def _infer_tracking(series, envelope_quantile,
             median_min_fraction
         )
         if system_tracking is Tracker.FIXED:
-            quadratic_median = _fit.quadratic(x=median.index, y=median)
+            quadratic_median = _fit.quadratic_r2(x=median.index, y=median)
             if quadratic_median < median_r2_min:
                 return Tracker.UNKNOWN
         elif system_tracking is Tracker.TRACKING:
-            quartic_median = _fit.quartic_restricted(
+            quartic_median = _fit.quartic_restricted_r2(
                 x=median.index,
                 y=median,
                 noon=middle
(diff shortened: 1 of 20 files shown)


wfvining:
Might want to rename `system.orientation` to `system.orientation_envelope` to highlight the specific method and leave room to add other methods in the future.

cwhanse:
I think we want a different name than `orientation` which is usually understood to mean azimuth and tilt, rather than fixed or tracking. I don't have a suggestion yet.

bt-:
> I think we want a different name than `orientation` which is usually understood to mean azimuth and tilt, rather than fixed or tracking. I don't have a suggestion yet.

I agree calling this `orientation` would be confusing to me. I'm not aware that there is a commonly accepted noun that encompasses- "is the system using fixed tilt racking or a tracker".  For a method that returns the type of racking, maybe something like `get_racking_type` would be intuitive?

wfvining:
I'm not sure about 'racking' either. In pvlib-python `racking_model` is used in the `PVSystem` constructor to mean something else. The original PVFleets QA code uses two names for this parameter, 'orientation' and 'mount'. Maybe `system.mount()`?

cwhanse:
or "is_tracking"?

wfvining:
I'm going to mark this ready for review. I think it might need more than one round, but I would like to start a discussion on some of the internals. Specifically:
* does the `fit_params` parameter make sense (is there a better way to pass these parameters)?
* does the winter/summer division make sense? Should we let users pass their own set of winter/summer months (probably, but I'm starting to feel like there are already a few too many parameters)?
* should we do a more sophisticated check on whether there is 'sufficient' data in each half of the year? The current scheme would allow data with even one point to proceed, probably resulting in a poor fit and misclassification.

pvanalytics/system.py
```diff
@@ -1 +1,199 @@
...
+            _group.by_minute(series).median(),
+            0.025
+        )
+        if system_tracking is Tracker.FIXED:
+            quadratic_median = _fit.quadratic(median)
+            if quadratic_median < 0.9:
+                return Tracker.UNKNOWN
+        elif system_tracking is Tracker.TRACKING:
+            quartic_median = _fit.quartic_restricted(median, middle)
+            if quartic_median < 0.9:
```
wfvining, line 104: magic number

pvanalytics/tests/util/test_fit.py
```diff
@@ -1,9 +1,9 @@
...
 
 
-def test_fit_not_datetime_raises_error(quadratic):
-    """Trying to fit to an integer index raises a ValueError."""
-    with pytest.raises(ValueError):
-        _fit.quadratic(quadratic)
+# def test_fit_not_datetime_raises_error(quadratic):
+#     """Trying to fit to an integer index raises a ValueError."""
+#     with pytest.raises(ValueError):
+#         _fit.quadratic(quadratic)
```
wfvining, line 9: Remove this entirely. (Can remove entire module.)

pvanalytics/system.py
```diff
@@ -1 +1,245 @@
...
+    whether the system is tracking or fixed. Finally an additional fit
+    is performed on the median of the data at each minute to determine
+    if there is substantial data below the envelope that does not
+    match the same profile. If the quadratic has a sufficiently good
+    fit then :py:const:`Tracker.FIXED` is returned. If the quartic has
+    a sufficiently good fit and the quadratic has a sufficiently bad
+    fit then :py:const:`Tracker.TRACKING` is returned. If neiher curve
+    fits well or there is a mismatch between the fit to the upper
+    envelope and the fit to the median then
+    :py:const:`Tracker.UNKNOWN` is returned.
```
cwhanse, line 131:
```suggestion
    whether the system is tracking or fixed.
    
    If the quadratic fit is a sufficiently good, then :py:const:`Tracker.FIXED` is returned.
    
    If the quartic fit is sufficiently good fit and the quadratic fit is sufficiently bad,
    then :py:const:`Tracker.TRACKING` is returned.
    
    If neither fit is sufficiently good then :py:const:`Tracker.UNKNOWN` is returned.

    Optionally, an additional fit is made to the median of the
    data at each minute to confirm the determination of tracking
    or fixed. If performed, tracking or fixed is judged from the fit
    to the median and this result must be consistent with the fit
    to the upper envelope. If not, :py:const:`Tracker.UNKNOWN`
    is returned.
```

```suggestion
    whether the system is tracking or fixed.
    

    ```

File: pvanalytics/system.py [hunk 4]
```diff
@@ -1 +1,245 @@
...
+    fit_params = fit_params or PVFLEETS_FIT_PARAMS
+    series_daytime = series[daytime]
+    clip_percent = (clipping[daytime].sum() / len(clipping[daytime])) * 100
+    march_september = series_daytime[
+        series_daytime.index.month.isin([5, 6, 7, 8])
+    ]
+    october_february = series_daytime[
+        series_daytime.index.month.isin([11, 12, 1, 2])
+    ]
+    if len(october_february) == 0 or len(march_september) == 0:
```
cwhanse, line 209: This seasonal logic needs to be described in the docstring. It is clearly meant for the northern hemisphere. We might want to provide kwargs to control it: `winter_months`, `seasonal_split`, I am just tossing ideas out.

pvanalytics/system.py
```diff
@@ -1 +1,245 @@
...
+        If None :py:data:`PVFLEETS_FIT_PARAMS` is used.
+    envelope_min_fraction : float, default 0.05
+        After calucluating the power or irradiance envelope data less
+        than `envelope_min_fraction` times the maximum of the envelope
+        is removed. This excludes data from morning and evening that
+        may interfere with curve fitting.
+    median_min_fraction : float, default 0.025
+        After calucluating the median power or irradiance at each
+        minute data less than `median_min_fraction` times the maximum
+        is removed. This excludes data from morning and evening that
```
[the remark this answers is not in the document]
wfvining: The maximum of the entire period. The flow is `99.5% quantile at each minute (upper enveolop)` -> `maximum of upper envelope` -> `remove data less than median_min_fraction * maximum`

pvanalytics/system.py
```diff
@@ -1 +1,245 @@
...
+        Timezone localized Series of power or irradiance data.
+    daytime : Series
+        Boolean Series with True for times that are during the day.
+    clipping : Series
+        Boolean Series identifying where power or irradiance is being
+        clipped.
+    clip_max : float, default 10.0
+        If the percent of data flagged as clipped is greater than
+        `clip_max` then tracking cannot be determined and
+        :py:const:`Tracker.UNKNOWN` is returned.
```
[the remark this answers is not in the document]
wfvining: If there is too much tracking then *neither* FIXED or TRACKING can be determined. Do you think the description should be rewritten to make that a little more clear? I could see "tracking cannot be determined" being misinterpreted.

```suggestion
    clip_max : float, default 10.0
        If the percent of data flagged as clipped is greater than
        `clip_max` then it cannot be determined whether the system 
        is equipped with a tracker and :py:const:`Tracker.UNKNOWN` 
        is returned.
```

pvanalytics/system.py
wfvining: I think `seasonal_split` with `([5, 6, 7, 8], [11, 12, 1, 2])` as a default could work. I would like to shy away from words like `winter` or `summer` to at least minimize the northern hemisphere bias.

pvanalytics/system.py
```diff
@@ -1,10 +1,283 @@
...
+                         envelope_quantile=0.995, fit_median=True,
+                         median_r2_min=0.9, fit_params=None,
+                         envelope_min_fraction=0.05,
+                         median_min_fraction=0.025,
+                         seasonal_split=None):
+    """Infer whether the system is equipped with a tracker.
+
+    Data is grouped by minute of the day and a maximum power or
+    irradiance envelope (the `envelope_quantile` value at each minute)
+    is calculated. Quadratic and quartic curves are fit to this daily
```
cwhanse, line 137:
```suggestion
    Data is grouped by season (optional) and within each season by the
    minute of the day. A maximum power or
    irradiance envelope (the `envelope_quantile` value at each minute)
    is calculated. Quadratic and quartic curves are fit to this daily
```

pvanalytics/system.py
```diff
@@ -1,10 +1,283 @@
...
+    envelope and the :math:`r^2` of the curve fits are used determine
+    whether the system is tracking or fixed.
+
+    If the quadratic fit is a sufficiently good, then
+    :py:const:`Tracker.FIXED` is returned.
+
+    If the quartic fit is sufficiently good fit and the quadratic fit
```
cwhanse, line 144:
```suggestion
    If, in both seasons, the quartic fit is sufficiently good and the quadratic fit
```

pvanalytics/system.py
```diff
@@ -1,10 +1,283 @@
...
+    returned.
+
+    If neither fit is sufficiently good then
+    :py:const:`Tracker.UNKNOWN` is returned.
+
+    Optionally, an additional fit is made to the median of the
+    data at each minute to confirm the determination of tracking
+    or fixed. If performed, tracking or fixed is judged from the fit
+    to the median and this result must be consistent with the fit
+    to the upper envelope. If not, :py:const:`Tracker.UNKNOWN`
```
cwhanse, line 155:
```suggestion
    or fixed. If performed, this result must be consistent with the fit
    to the upper envelope. If not, :py:const:`Tracker.UNKNOWN`
```

pvanalytics/system.py
```diff
@@ -1,10 +1,283 @@
...
+        the associated fit parameters are applicable. The values of
+        the dictionary are themselves dictionaries with keys
+        ``'fixed'`` and ``'tracking'``, which give the minimum
+        :math:`r^2` for the curve fits, and ``'fixed_max'`` which
+        gives the maximum :math:`r^2` for a quadratic fit if the
+        system appears to have a tracker.
+
+        If None :py:data:`PVFLEETS_FIT_PARAMS` is used.
+    envelope_min_fraction : float, default 0.05
+        After calucluating the power or irradiance envelope data less
```
cwhanse, line 197:
```suggestion
        After calculating the power or irradiance envelope, data less
```

pvanalytics/system.py
```diff
@@ -1,10 +1,283 @@
...
+        Quantile used to determine the upper power or irradiance
+        envelope.
+    fit_median : boolean, default True
+        Perform a secondary fit with the median power or irradiance to
+        validate that the profile is consistent through the entire
+        data set.
+    median_r2_min : float, default 0.9
+        Minimum :math:`r^2` for a curve fit to the median power or
+        irradiance at each minute of the day (Applies only if
+        `fit_median` is True).
```
cwhanse, line 181: Editorial preference: group the arguments related to the median fit together, after the arguments for the envelope fit.

pvanalytics/system.py
```diff
@@ -1,10 +1,283 @@
...
+    seasonal_split : tuple of list of int, optional
+        Tuple specifying a set of winter months and a set of summer
+        months. The order is not important. The months are specified
+        as integers between 1 and 12. If not specified defaults to
+        ([5, 6, 7, 8], [11, 12, 1, 2]) which works well for North
+        America. Data is split in to two groups, one for each set and
+        curve fits are applied to the upper envelope of each group
+        independently. If the curve fits produce different results
+        (e.g. one TRACKING and one FIXED) then
+        :py:const:`Tracker.UNKNOWN` is returned.
```
cwhanse, line 215: This one is a bit confusing. Defaults to None but None signals a different default. So it appears that None doesn't turn the seasonal split off, it's a shortcut to the actual default.

Is there a way to turn seasonal splitting off? If yes, then leave the default at None, and describe ([5, ...]) etc. as values that work well in the northern hemisphere.  Otherwise I think it would be better to default to the tuple in the comment and put that default value in the docstring in place of 'optional'.

If we have an option to turn seasonal splitting off, it should be tested.

pvanalytics/system.py
```diff
@@ -1,10 +1,283 @@
...
+    """Infer whether the system is equipped with a tracker.
+
+    Data is grouped by minute of the day and a maximum power or
+    irradiance envelope (the `envelope_quantile` value at each minute)
+    is calculated. Quadratic and quartic curves are fit to this daily
+    envelope and the :math:`r^2` of the curve fits are used determine
+    whether the system is tracking or fixed.
+
+    If the quadratic fit is a sufficiently good, then
+    :py:const:`Tracker.FIXED` is returned.
```
cwhanse, line 142:
```suggestion
    If the quadratic fit is a sufficiently good in both seasons, then
    :py:const:`Tracker.FIXED` is returned.
```

pvanalytics/system.py
```diff
@@ -1,10 +1,283 @@
...
+
+    If the quartic fit is sufficiently good fit and the quadratic fit
+    is sufficiently bad, then :py:const:`Tracker.TRACKING` is
+    returned.
+
+    If neither fit is sufficiently good then
+    :py:const:`Tracker.UNKNOWN` is returned.
```
cwhanse, line 149:
```suggestion
    If neither fit is sufficiently good, or the results from each season disagree, then
    :py:const:`Tracker.UNKNOWN` is returned.
```

pvanalytics/tests/test_system.py
```diff
@@ -1,11 +1,229 @@
...
+        system,
+        albuquerque,
+        orientation_strategy='south_at_latitude_tilt'
+    )
+    mc.run_model(summer_clearsky)
+    return mc.ac
+
+
+def test_ghi_tracking_envelope_fixed(summer_ghi):
+    """Clearsky GHI for has a FIXED Tracker."""
```
cwhanse, line 78:
```suggestion
    """Clearsky GHI for a system that is FIXED."""
```

pvanalytics/tests/test_system.py
```diff
@@ -1,11 +1,229 @@
...
+    """Simulated system under clearsky condidtions is FIXED."""
+    assert system.is_tracking_envelope(
+        summer_power_fixed,
+        summer_power_fixed > 0,
+        pd.Series(False, index=summer_power_fixed.index)
+    ) is system.Tracker.FIXED
+
+
+def test_power_tracking_envelope_tracking(summer_power_tracking):
+    """Simulated single axis tracker is identifified as TRACKING."""
```
cwhanse, line 96:
```suggestion
    """Simulated single axis tracker is identified as TRACKING."""
```

pvanalytics/system.py
wfvining: @cwhanse I changed the docstring, but I think we should leave the actual default value as `None` to avoid problems with mutable defaults.

kanderso-nrel: How about tuple of tuple instead of tuple of list for mutability concerns?

wfvining: I think I'm unclear on the immutability of a tuple when it holds mutable data. While you can't assign a new object to the elements of a tuple you can still mutate the objects that it already contains
```python
x = ([1, 2], [3])
x[1].append(4)
```
results in `x = ([1, 2], [3, 4])`.

As far as I can tell, that results in the same problem as having a bare list as the default.

kanderso-nrel: After reading it again, my previous comment reminds me of the [buffalo sentence](https://en.wikipedia.org/wiki/Buffalo_buffalo_Buffalo_buffalo_buffalo_buffalo_Buffalo_buffalo).

To clarify, I mean using `((5, 6, 7, 8), (11, 12, 1, 2))` as the default value instead of `([5, 6, 7, 8], [11, 12, 1, 2])`.

wfvining: My mistake. that would work! Thanks for your advice, @kanderso-nrel, I'm not used to thinking about tuples in the laid-back way that python treats them.

cwhanse: I'm not used to parsing the buffalo sentence....

Thanks @kanderso-nrel for the help and the Friday distraction :)

pvanalytics/system.py
```diff
@@ -1,10 +1,280 @@
...
+        ``'fixed'`` and ``'tracking'``, which give the minimum
+        :math:`r^2` for the curve fits, and ``'fixed_max'`` which
+        gives the maximum :math:`r^2` for a quadratic fit if the
+        system appears to have a tracker.
+
+        If None :py:data:`PVFLEETS_FIT_PARAMS` is used.
+    seasonal_split : tuple of list-like, default ((5,6,7,8), (11,12,1,2))
+        Two-tuple specifying a set of winter months and a set of summer
+        months. The order is not important. The months are specified
+        as integers between 1 and 12. The default value works well for North
```
cwhanse, line 208:
```suggestion
    seasonal_split : tuple of list-like, default ((5,6,7,8), (11,12,1,2))
        Two-tuple specifying tuples of months for summer and winter seasons.
        The months are specified
        as integers between 1 and 12. The default value works well for North
```

I hate to churn on this but I think it's still less clear than it should be. "The order is not important" true in that the code doesn't care which tuple is considered summer, but I think it would help users to identify one element as summer and the other as winter.

Pushing that a bit further, what if the default was a dict instead? `{'summer': (5, 6, 7, 8), 'winter': (11, 12, 1, 2)}`

wfvining: I totally agree. I was just looking at that language. You make a good point about being "explicit" about what is winter and what is summer. Rather than a dictionary, what about two parameters? `winter=(11, 12, 1, 2)` and `summer = (5, 6, 7, 8)` ?

Might want to change the semantics a bit too, and allow `None` for either or both seasons. I will need to write some additional tests in that case, but the more I think about it the more it seems like the right thing to do.

cwhanse: I'm OK with two parameters, but I think it should be easy to disable this optional check which would be an advantage of a single argument (pass `None` to skip)

wfvining:
@cwhanse I implemented the optional seasonal split with the dictionary parameter discussed above (along with new tests). Would you mind taking a quick look to see if you think it is reasonable? I didn't want to put a dictionary as the default value directly (mutability concerns) so I used a descriptive key-word.

I also added a warning for possible user error when neither season has data (seems more polite than a ValueError, or silently returning UNKNOWN).

pvanalytics/system.py
```diff
@@ -1,10 +1,289 @@
...
+            median_min_fraction
+        )
+    summer = series_daytime[
+        series_daytime.index.month.isin(seasonal_split.get('summer') or [])
+    ]
+    winter = series_daytime[
+        series_daytime.index.month.isin(seasonal_split.get('winter') or [])
+    ]
+    if len(winter) == 0 and len(summer) == 0:
+        warnings.warn("no data in both winter and summer months")
```
cwhanse, line 274:
```suggestion
        warnings.warn("no data in winter or summer months")
```

cwhanse:
Looks good you decide about that last comment.

pvanalytics/system.py
wfvining: I thought this could be misleading, since the warning is only issued if both seasons are empty.

cwhanse: 'both winter and summer' made me think the code was expecting a subset of the data to be in both months.

cwhanse: maybe "data is required in both winter and summer months"?

wfvining: "unknown tracking - both winter and summer months have no data"?

cwhanse: better - I'm reacting to the g[rammar error](https://dictionary.cambridge.org/us/grammar/british-grammar/neither-neither-nor-and-not-either_2) when using "...both... not..." 

"Unknown tracking - neither winter nor summer months have data"

wfvining:
```suggestion
        warnings.warn("Unknown tracking - neither winter nor summer months have data.")
```

pvanalytics/system.py
```diff
@@ -1,10 +1,289 @@
...
+            median_min_fraction
+        )
+    summer = series_daytime[
+        series_daytime.index.month.isin(seasonal_split.get('summer') or [])
+    ]
+    winter = series_daytime[
+        series_daytime.index.month.isin(seasonal_split.get('winter') or [])
+    ]
+    if len(winter) == 0 and len(summer) == 0:
+        warnings.warn("Unknown tracking - neither winter nor summer months have data.")
```
wfvining, line 274:
```suggestion
        warnings.warn(
            "Unknown tracking - neither winter nor summer months have data."
        )
```

PR #52, opened 2020-06-09
Identify days where system output looks like tracking or fixed

### LINKED ISSUE #51: Detect sunny days
The PVFleets QA project does this by fitting curves (quadratic for fixed systems, quartic for tracking systems) on each day.

wfvining:
Introduces two functions that fit curves to power or irradiance data for each day to determine whether that day matches the expected profile of a fixed or tracking system (parabolic or quartic respectively).

Closes #51


pvanalytics/features/orientation.py
```diff
@@ -0,0 +1,185 @@
...
+    marked False.
+
+    Parameters
+    ----------
+    power_or_irradiance : Series
+        Timezone localized series of power or irradiance measurements.
+    daytime : Series
+        Boolean series with True for times that are during the
+        day. For best results this mask should exclude early morning
+        and evening as well as night. Morning and evening may have
```
cwhanse, line 73:
```suggestion
        and late afternoon as well as night. These times may have
```

pvanalytics/features/orientation.py
```diff
@@ -0,0 +1,185 @@
...
+        problems with shadows that interfere with curve fitting, but
+        do not necessarily indicate that the day was not sunny.
```
cwhanse, line 75: confused by the reference to "sunny" days here - is the returned series True only when both sunny and looks-like-tracking?

pvanalytics/features/orientation.py
```diff
@@ -0,0 +1,185 @@
...
+        Maximum :math:`r^2` for a quadratic fit, if the quadratic fit
+        is better than this, then tracking/fixed cannot be determined
+        and the day is marked False.
+    min_hours : float, default 5.0
+        Minimum number of hours with data to attempt a fit on a day.
+    peak_min : float, default None
+        The maximum value for a day must be greater than `peak_min`
+        for a fit to be attempted. If the maximum for a day is less
+        than `peak_min` then the day will be marked False.
+    late_morning : str, default '08:45'
```
cwhanse, line 88: Not to complicate matters, but there are different ways to do this screening - on solar zenith for example. Maybe replace `late_morning` and `early_afternoon` with another Boolean series, a mask which is True where data is considered in the check for stuck trackers.

On that topic, I don't see where the check for stuck trackers is done. It looks like the morning/afternoon window is used to check the data for the appearance of a fixed tilt system, rather than a tracker that is mispointed and not moving (which more than likely would not have a quadratic shape). Maybe revise the comments and docstring so it doesn't appear that this function is detecting stuck trackers.

pvanalytics/features/orientation.py
```diff
@@ -0,0 +1,185 @@
...
+    #
+    # The date needs to be localized in the same timezone as
+    # data.index so we can easily reindex the series with the original
+    # index at a later time.
+    return data.groupby(
+        pd.to_datetime(data.index.date).tz_localize(data.index.tz)
+    )
+
+
+def tracking(power_or_irradiance, daytime, correlation_min=0.94,
```
cwhanse, line 53: Let's name this function after NREL, to make space for alternate methods. `tracking_nrel` and let Matt change it.

pvanalytics/features/orientation.py
```diff
@@ -0,0 +1,185 @@
...
+def _freqstr_to_hours(freq):
+    # Convert pandas freqstr to hours (as a float)
+    if freq.isalpha():
+        freq = '1' + freq
+    return pd.to_timedelta(freq).seconds / 60
+
+
+def _hours(data, freq):
+    # Return the number of hours in `data` with data where timestamp
+    # spacing is given by `freq`.
```
cwhanse, line 31:
```suggestion
    # Return the number of hours in `data` with timestamp
    # spacing given by `freq`.
```

pvanalytics/tests/features/test_orientation.py
```diff
@@ -0,0 +1,137 @@
...
+        'module_parameters': module,
+        'inverter_parameters': inverter,
+        'temperature_model_parameters': temperature_model_parameters
+    }
+
+
+@pytest.fixture(scope='module')
+def clearsky(times, albuquerque):
+    """Clearsky at `times` in `albuquerque`."""
+    return albuquerque.get_clearsky(times)
```
cwhanse, line 53:
```suggestion
    return albuquerque.get_clearsky(times, model='haurwitz')
```
would avoid needing to import `tables`, if this calculation is why `tables` is needed

pvanalytics/features/orientation.py
wfvining: I will change these parameters to another boolean series since it does add substantial flexibility. I kept the original behavior because I was not sure how to assign a reasonable default and I thought it might be cumbersome to always pass two day/night masks. Maybe the 'midday' mask should default to `None` and in that case `daytime` is used.

> It looks like the morning/afternoon window is used to check the data for the appearance of a fixed tilt system, rather than a tracker that is mispointed and not moving (which more than likely would not have a quadratic shape)

My understanding from the comments in the original Fleets implementation is that aggressively chopping off the morning and afternoon would result in something close enough to quadratic that the curve fitting would be somewhat successful.

Then the three criteria for whether the system is tracking come in to play:
1. quartic has a sufficiently good fit
2. quadratic has a sufficiently bad fit
3. quartic has a better fit than the quadratic

I believe that in this case it would be the 'quartic fits better than quadratic' criteria (3) that fails, rather than the 'quadratic has a sufficiently bad fit' criteria (2). One of the key points is that the quartic that we fit to the data is heavily restricted to prevent it from fitting well to anything other than the profile of a tracking PV system.

I take your point though. This is more of a test that the curve does not look like a functioning tracker than that the curve looks like a stuck tracker. I will clarify that in the documentation, and I think I should add some of the content from this comment to the documentation as well.

.github/workflows/lint-and-test.yml
```diff
@@ -22,6 +22,18 @@ jobs:
       run: |
         python -m pip install --upgrade pip
         pip install -r requirements.txt
+    # This is a bit of a hack to make the Github runners work.
+    #
+    # The latest version of tables (3.6) fails to install with python
+    # 3.5 and the older version of tables fails to install with python
+    # 3.8.
+    - name: Install old tables
```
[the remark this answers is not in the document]
wfvining: Yes. I saw your comment below. Thanks for the pointer, I will change and revert the changes to the CI builds.

pvanalytics/features/orientation.py
wfvining, line 75:
```suggestion
        do not necessarily indicate that the tracker is not working.
```

pvanalytics/tests/features/test_orientation.py
wfvining: Looks like 'haurwitz' won't actually work for these tests. `ModelChain.run_model` expects weather to have 'ghi', 'dhi', and 'dni'. When `model='haurwitz'`, `pvlib.Location.get_clearsky()` only returns 'ghi'.

Does`method= 'simplified_solis'` require tables?

cwhanse: > Doesmodel='simplified_solis' require tables?

It doesn't, and would serve the purpose here.

docs/api.rst
```diff
@@ -141,6 +141,27 @@ Clearsky
 
    features.clearsky.reno
 
+Orientation
+-----------
+
+The following funcitons examine power or irradiance data to identify
```
cwhanse, line 147:
```suggestion
The following functions examine power or irradiance data to identify
```

pvanalytics/features/orientation.py
```diff
@@ -0,0 +1,193 @@
...
+                  fixed_max=0.96, min_hours=5, peak_min=None,
+                  midday=None):
+    """Flag days that match the profile of a single-axis tracking PV system.
+
+    For the values on a day to be marked True, they must satisfy the
+    following three conditions:
+
+    1. a restricted quartic [#]_ must fit the data with :math:`r^2`
+       greater than `correlation_min`
+    2. the :math:`r^2` for a quadratic fit must be less than
```
cwhanse, line 63: I would switch the order of the 2nd and 3rd criteria

pvanalytics/features/orientation.py
```diff
@@ -0,0 +1,193 @@
...
+       15% of the median for the day, and it must open downwards.
+
+    Parameters
+    ----------
+    power_or_irradiance : Series
+        Timezone localized series of power or irradiance measurements.
+    daytime : Series
+        Boolean series with True for times that are during the
+        day. For best results this mask should exclude early morning
+        and late afternoon as well as night. These times may have
```
cwhanse, line 82:
```suggestion
        and late afternoon as well as night. Data at these times may have
```

pvanalytics/features/orientation.py
```diff
@@ -0,0 +1,193 @@
...
+        problems with shadows that interfere with curve fitting, but
+        do not necessarily indicate that the tracker is not working.
```
cwhanse, line 84:
```suggestion
        problems with shadows that interfere with curve fitting.
```

pvanalytics/features/orientation.py
```diff
@@ -0,0 +1,193 @@
...
+    correlation_min : float, default 0.94
```
cwhanse, line 85:
```suggestion
    r2_min : float, default 0.94
```

R2 is the coefficient of determination (R is the correlation coefficient).

pvanalytics/features/orientation.py
```diff
@@ -0,0 +1,193 @@
...
+        Maximum :math:`r^2` for a quadratic fit, if the quadratic fit
+        is better than this, then tracking/fixed cannot be determined
+        and the day is marked False.
+    min_hours : float, default 5.0
+        Minimum number of hours with data to attempt a fit on a day.
+    peak_min : float, default None
+        The maximum value for a day must be greater than `peak_min`
+        for a fit to be attempted. If the maximum for a day is less
+        than `peak_min` then the day will be marked False.
+    midday : Series, default None
```
cwhanse, line 97: Let's combine `daytime` and `midday` into a single mask that selects data to be used in the fitting. The user will need to combine the day/night filter with any time/zenith limits outside this function.

I see your earlier point about two masks being awkward.

pvanalytics/features/orientation.py
```diff
@@ -0,0 +1,193 @@
...
+        Timezone localized series of power or irradiance measurements.
+    daytime : Series
+        Boolean series with True for times that are during the
+        day. For best results this mask should exclude early morning
+        and late afternoon as well as night. These times may have
+        problems with shadows that interfere with curve fitting, but
+        do not necessarily indicate that the tracker is not working.
+    correlation_min : float, default 0.94
+        Minimum :math:`r^2` for a day to be considered sunny.
+    fixed_max : float, default 0.96
```
cwhanse, line 87: I'm struggling with the name here. It should reference `r2` somehow. `r2_max_fixed` maybe? Or just `r2_max`?

The description seems a bit tortuous, what about : "If the :math:`r^2` of the quadratic fit exceeds (parameter), then tracking/fixed cannot be distinguished and the day is marked False."

pvanalytics/features/orientation.py
```diff
@@ -0,0 +1,193 @@
...
+        & (fixed_days < fixed_max)
+    ).reindex(power_or_irradiance.index, method='pad', fill_value=False)
+
+
+def fixed_nrel(power_or_irradiance, daytime, correlation_min=0.94,
+               min_hours=5, peak_min=None):
+    """Flag days that match the profile of a fixed PV system.
+
+    Fixed days are identified when the :math:`r^2` for a quadratic fit
+    to the power data is greater than `correlation_min`.
```
cwhanse, line 147:
```suggestion
    to the power data is greater than `r2_min`.
```

pvanalytics/features/orientation.py
```diff
@@ -0,0 +1,193 @@
...
+        day. For best results this mask should exclude early morning
+        and evening as well as night. Morning and evening may have
+        problems with shadows that interfere with curve fitting, but
+        do not necessarily indicate that the day was not sunny.
+    correlation_min : float, default 0.94
+        Minimum :math:`r^2` for a day to be considered sunny.
+    min_hours : float, default 5.0
+        Minimum number of hours with data to attempt a fit on a day.
+    peak_min : float, default None
+        The maximum value for a day must be greater than `peak_min`
```
cwhanse, line 164:
```suggestion
        The maximum `power_or_irradiance` value for a day must be greater than `peak_min`
```

pvanalytics/features/orientation.py
```diff
@@ -0,0 +1,193 @@
...
+    correlation_min : float, default 0.94
+        Minimum :math:`r^2` for a day to be considered sunny.
+    fixed_max : float, default 0.96
+        Maximum :math:`r^2` for a quadratic fit, if the quadratic fit
+        is better than this, then tracking/fixed cannot be determined
+        and the day is marked False.
+    min_hours : float, default 5.0
+        Minimum number of hours with data to attempt a fit on a day.
+    peak_min : float, default None
+        The maximum value for a day must be greater than `peak_min`
```
cwhanse, line 94:
```suggestion
        The maximum `power_or_irradiance` value for a day must be greater than `peak_min`
```

pvanalytics/features/orientation.py
wfvining: I'd like to talk this through a bit with @matt14muller. I think that in order to get a reasonably good fit with the particular quartic that he devised it is important to include some of the morning and evening. In my testing, I found that if we were to decrease the maximum zenith angle to 85 degrees the r^2 for the quartic drops to ~0.8 for the tracking system used in the tests. 

Since the point of the `midday` mask is to eliminate the long shadowed periods that are only present when the tracker is not functioning correctly I'm not sure that combining them is the right thing to do. It is a bit awkward, but - at the moment - I think the awkwardness is acceptable since it keeps parameters with different semantics separate. I'm happier with the `None` default than I thought I would be.

pvanalytics/features/orientation.py
```diff
@@ -0,0 +1,193 @@
...
+
+    For the values on a day to be marked True, they must satisfy the
+    following three conditions:
+
+    1. a restricted quartic [#]_ must fit the data with :math:`r^2`
+       greater than `correlation_min`
+    2. the :math:`r^2` for a quadratic fit must be less than
+       `fixed_max`
+    3. the :math:`r^2` for the restricted quartic fit must be greater
+       than the :math:`r^2` for a quadratic fit
```
wfvining, line 66:
```suggestion
    2. the :math:`r^2` for the restricted quartic fit must be greater
       than the :math:`r^2` for a quadratic fit
    3. the :math:`r^2` for a quadratic fit must be less than
       `fixed_max`
```

wfvining:
Wondering if `orientation.tracking` should be reworked to return three Series:
* `tracking` True where the profile matches that of a tracker on a clear day (good quartic fit)
* `stuck` True where the profile matches a stuck tracker (i.e. vaguely quadratic)
* `unknown` True where a tracking/stuck determination cannot be made

pvanalytics/features/orientation.py
wfvining: I think this has been addressed by renaming the parameter and clarifying its purpose in the documentation.

pvanalytics/features/orientation.py
```diff
@@ -0,0 +1,175 @@
...
+    return data.count() * _freqstr_to_hours(freq)
+
+
+def tracking_nrel(power_or_irradiance, daytime, r2_min=0.915,
+                  r2_fixed_max=0.96, min_hours=5, peak_min=None,
+                  quadratic_mask=None):
+    """Flag days that match the profile of a single-axis tracking PV system.
+
+    For the values on a day to be marked True, they must satisfy the
+    following three conditions:
```
cwhanse, line 41:
```suggestion
```

pvanalytics/features/orientation.py
```diff
@@ -0,0 +1,175 @@
...
+    ----------
+    power_or_irradiance : Series
+        Timezone localized series of power or irradiance measurements.
+    daytime : Series
+        Boolean series with True for times that are during the
+        day. For best results this mask should exclude early morning
+        and late afternoon as well as night. Data at these times may have
+        problems with shadows that interfere with curve fitting.
+    r2_min : float, default 0.915
+        Minimum :math:`r^2` for a day to be considered sunny.
```
cwhanse, line 67:
```suggestion
        Minimum :math:`r^2` of a quartic fit for a day to be marked True.
```

pvanalytics/features/orientation.py
```diff
@@ -0,0 +1,175 @@
...
+    peak_min : float, default None
+        The maximum `power_or_irradiance` value for a day must be
+        greater than `peak_min` for a fit to be attempted. If the
+        maximum for a day is less than `peak_min` then the day will be
+        marked False.
+    quadratic_mask : Series, default None
+        If None then `daytime` is used. This Series is used to remove
+        morning and afternoon times from the data before applying a
+        quadratic fit to check for a stuck tracker. It should
+        typically exclude more data than `daytime` in order to
```
cwhanse, line 83:
```suggestion
        quadratic fit. The mask should
        typically exclude more data than `daytime` in order to
```

pvanalytics/features/orientation.py
```diff
@@ -0,0 +1,175 @@
...
+        eliminate long tails in the morning or afternoon that appear
+        when a tracker is stuck in a West or East orientation.
```
cwhanse, line 85:
```suggestion
        eliminate long tails in the morning or afternoon that can
        appear if a tracker is stuck in a West or East orientation.
```

pvanalytics/features/orientation.py
```diff
@@ -0,0 +1,175 @@
...
+        & (fixed_days < r2_fixed_max)
+    ).reindex(power_or_irradiance.index, method='pad', fill_value=False)
+
+
+def fixed_nrel(power_or_irradiance, daytime, r2_min=0.94,
+               min_hours=5, peak_min=None):
+    """Flag days that match the profile of a fixed PV system.
+
+    Fixed days are identified when the :math:`r^2` for a quadratic fit
+    to the power data is greater than `r2_min`.
```
cwhanse, line 129:
```suggestion
    This algorithm relies on the observation that the power profile of a
    fixed tilt PV system often resembles a quadratic polynomial on a
    sunny day, with a single peak when the sun is near the system azimuth.
    
    A day is marked True when the :math:`r^2` for a quadratic fit to the
    power data is greater than `r2_min`.
```

pvanalytics/features/orientation.py
```diff
@@ -0,0 +1,175 @@
...
+def _hours(data, freq):
+    # Return the number of hours in `data` with timestamp
+    # spacing given by `freq`.
+    return data.count() * _freqstr_to_hours(freq)
+
+
+def tracking_nrel(power_or_irradiance, daytime, r2_min=0.915,
+                  r2_fixed_max=0.96, min_hours=5, peak_min=None,
+                  quadratic_mask=None):
+    """Flag days that match the profile of a single-axis tracking PV system.
```
cwhanse, line 38:
```suggestion
    """Flag days that match the profile of a single-axis tracking PV system
    on a sunny day.
```

pvanalytics/features/orientation.py
```diff
@@ -0,0 +1,175 @@
...
+    return (
+        (tracking_days > r2_min)
+        & (tracking_days > fixed_days)
+        & (fixed_days < r2_fixed_max)
+    ).reindex(power_or_irradiance.index, method='pad', fill_value=False)
+
+
+def fixed_nrel(power_or_irradiance, daytime, r2_min=0.94,
+               min_hours=5, peak_min=None):
+    """Flag days that match the profile of a fixed PV system.
```
cwhanse, line 126:
```suggestion
    """Flag days that match the profile of a fixed PV system on a sunny day.
```

pvanalytics/features/orientation.py
```diff
@@ -0,0 +1,175 @@
...
+    ----------
+    power_or_irradiance : Series
+        Timezone localized series of power or irradiance measurements.
+    daytime : Series
+        Boolean series with True for times that are during the
+        day. For best results this mask should exclude early morning
+        and evening as well as night. Data at these times may have
+        problems with shadows that interfere with curve fitting.
+    r2_min : float, default 0.94
+        Minimum :math:`r^2` for a day to be considered sunny.
```
cwhanse, line 141:
```suggestion
        Minimum :math:`r^2` of a quadratic fit for a day to be marked True.
```

pvanalytics/features/orientation.py
```diff
@@ -0,0 +1,175 @@
...
+        The maximum `power_or_irradiance` value for a day must be
+        greater than `peak_min` for a fit to be attempted. If the
+        maximum for a day is less than `peak_min` then the day will be
+        marked False.
+
+    Returns
+    -------
+    Series
+        True for values on days where `power_or_irradiance` matches
+        the expected parabolic profile for a fixed PV system.
```
cwhanse, line 154:
```suggestion
        the expected parabolic profile for a fixed PV system on a sunny day.
```

pvanalytics/tests/features/test_orientation.py
```diff
@@ -0,0 +1,159 @@
...
+
+
+@pytest.fixture(scope='module')
+def solarposition(times, albuquerque):
+    """Solar position at `times` in `albuquerque`."""
+    return albuquerque.get_solarposition(times)
+
+
+def test_clearsky_ghi_fixed(clearsky, solarposition):
+    """every day of clearsky GHI is a sunny day."""
```
cwhanse, line 63:
```suggestion
    """Identify every day as fixed, since clearsky GHI is sunny."""
```

docs/api.rst
```diff
@@ -141,6 +141,32 @@ Clearsky
...
+to identify stuck days. Combining these fixed days with tracking days
+using an exclusive or should give a reasonable estimate of relatively
+sunny days where the tracker is stuck.
+
+.. autosummary::
+   :toctree: generated/
+
+   features.orientation.fixed_nrel
+   features.orientation.tracking_nrel
+
```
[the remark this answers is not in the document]
wfvining: This is referred to in `quality.irradiance`  I think it needs to be kept at the bottom for Sphinx to render the docs correctly, but I'm not 100% sure. I'll try moving it up and see if we can keep it close to where it is referenced.

PR #55, opened 2020-06-22
Test correlation between module temperature and irradiance

### LINKED ISSUE #47: Add module-temperature/irradiance correlation check
PVFleets QA checks that the module temperature is correlated with the irradiance. This is a pass/fail quality check: if the temperature and irradiance have sufficiently high correlation over the entire series then the temperature sensor passes. It might be useful to apply the correlation check on rolling windows as well and return a boolean series rather than just a pass/fail.

I'm not sure where this should go in the `quality` module.

wfvining:
This is a pass/fail test for module temperature data. In the future we might add a rolling version that passes/fails particular values based on correlation in a rolling window, but for now we just pass/fail the entire temperature time series.

I put this in `weather` because that is where other temperature quality checks live, but I'm not 100% sure that this is the best place for it.

Closes #47


pvanalytics/quality/weather.py
```diff
@@ -97,3 +98,30 @@ def wind_limits(wind_speed, limits=(0.0, 50.0)):
...
+        quality (outliers and other problems removed).
+    correlation_min : float, default 0.5
+        Minimum correlation between `module_temperature` and
+        `irradiance` for the module temperature sensor to 'pass'
+
+    Returns
+    -------
+    bool
+        True if the `module_temperature` is correlated with
+        `irradiance`.
```
cwhanse, line 123:
```suggestion
        True if the correlation between `module_temperature` and
        `irradiance` exceeds `correlation_min`.
```

docs/api.rst
```diff
@@ -134,6 +134,20 @@ Quality checks for weather data.
    quality.weather.temperature_limits
    quality.weather.wind_limits
 
+In addition to validating temperature limits, module-temperature can
+be validated by testing whether it is correlated with irradiance. Poor
```
cwhanse, line 138:
```suggestion
In addition to validating temperature by comparing with limits, module 
temperature should be positively correlated with irradiance. Poor
```

PR #57, opened 2020-06-23
Daily insolation limits

wfvining:
Identify days where the total irradiance is too high or too low. This catches incorrectly labeled sensors (e.g. labeled GHI, but actually POA) or days with strange shifts (for example I saw a data set where the units change from Watts to Milliwatts for a week or two). Existing functions might also catch these, but the aggregate daily check seems like a worthwhile addition to me.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -43,7 +43,7 @@ eliminate data that is unrealistically high.
    quality.irradiance.clearsky_limits
 
 You may want to identify entire days that have unrealistically high or
-low irradiance. The following function examines daily insolation,
+low insolation. The following function examines daily insolation,
 validating that it is within a reasonable range of the expected
 clearsky insolation for the same day.
 
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -43,14 +43,14 @@ eliminate data that is unrealistically high.
    quality.irradiance.clearsky_limits
 
 You may want to identify entire days that have unrealistically high or
-low irradiance. The following function examines summed daily
-irradiance, validating that it is within a reasonable range of the
-summed clearsky irradiance for the same day.
+low irradiance. The following function examines daily insolation,
+validating that it is within a reasonable range of the expected
+clearsky insolation for the same day.
 
 .. autosummary::
    :toctree: generated/
 
-   quality.irradiance.daily_limits
+   quality.irradiance.daily_insolation_limits
 
 Gaps
 ----
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -42,6 +42,16 @@ eliminate data that is unrealistically high.
 
    quality.irradiance.clearsky_limits
 
+You may want to identify entire days that have unrealistically high or
+low irradiance. The following function examines summed daily
+irradiance, validating that it is within a reasonable range of the
+summed clearsky irradiance for the same day.
+
+.. autosummary::
+   :toctree: generated/
+
+   quality.irradiance.daily_limits
+
 Gaps
 ----
 
--- a/pvanalytics/quality/irradiance.py
+++ b/pvanalytics/quality/irradiance.py
@@ -417,10 +417,12 @@ def _daily_total(series):
             integrate.trapz,
             dx=freq_hours
         )
-    df = series.to_frame(name='value')
-    df['hour'] = (series.index.minute / 60) + series.index.hour
-    return df.resample('D').apply(
-        lambda day: integrate.trapz(y=day['value'], x=day['hour'])
+    hours = pd.Series(
+        (series.index.minute / 60) + series.index.hour,
+        index=series.index
+    )
+    return series.resample('D').apply(
+        lambda day: integrate.trapz(y=day, x=hours[day.index])
     )
 
 
--- a/pvanalytics/quality/irradiance.py
+++ b/pvanalytics/quality/irradiance.py
@@ -400,8 +400,14 @@ def _to_hours(freqstr):
 
 
 def _daily_total(series):
+    # Resample the series returning the integral for each day.
+    #
+    # Assumes that timestamps are evenly spaced and uses the Trapezoid
+    # rule for integration.
     freq_hours = _to_hours(pd.infer_freq(series.index))
-    return series.resample('D').sum() * freq_hours
+    return series.resample('D').apply(
+        lambda day: freq_hours * (day[:-1] + day[1:]).sum()
+    )
 
 
 def daily_limits(irrad, clearsky, daily_min=0.4, daily_max=1.25):
--- a/pvanalytics/quality/irradiance.py
+++ b/pvanalytics/quality/irradiance.py
@@ -410,7 +410,7 @@ def _daily_total(series):
     )
 
 
-def daily_limits(irrad, clearsky, daily_min=0.4, daily_max=1.25):
+def daily_insolation_limits(irrad, clearsky, daily_min=0.4, daily_max=1.25):
     """Check that daily insolation lies between minimum and maximum values.
 
     Irradiance measurements and clear-sky irradiance on each day are
--- a/pvanalytics/quality/irradiance.py
+++ b/pvanalytics/quality/irradiance.py
@@ -407,8 +407,8 @@ def _daily_total(series):
 def daily_limits(irrad, clearsky, daily_min=0.4, daily_max=1.25):
     """Check that daily insolation lies between minimum and maximum values.
 
-    Irradiance measurements and clear-sky irradiance on each day are integrated with the
-    left-hand rule to calculate daily insolation.
+    Irradiance measurements and clear-sky irradiance on each day are
+    integrated with the left-hand rule to calculate daily insolation.
 
     .. note::
 
@@ -432,8 +432,8 @@ def daily_limits(irrad, clearsky, daily_min=0.4, daily_max=1.25):
     Returns
     -------
     Series
-        True for values on days where the ratio of daily insolation to daily clearsky insolation is between `daily_min`
-        and `daily_max`.
+        True for values on days where the ratio of daily insolation to
+        daily clearsky insolation is between `daily_min` and `daily_max`.
 
     Notes
     -----
--- a/pvanalytics/quality/irradiance.py
+++ b/pvanalytics/quality/irradiance.py
@@ -442,9 +442,9 @@ def daily_limits(irrad, clearsky, daily_min=0.4, daily_max=1.25):
 
     """
     daily_irradiance = _daily_total(irrad)
-    daily_cleasky = _daily_total(clearsky)
+    daily_clearsky = _daily_total(clearsky)
     good_days = util.check_limits(
-        daily_irradiance/daily_cleasky,
+        daily_irradiance/daily_clearsky,
         upper_bound=daily_max,
         lower_bound=daily_min
     )
--- a/pvanalytics/quality/irradiance.py
+++ b/pvanalytics/quality/irradiance.py
@@ -451,13 +451,13 @@ def daily_insolation_limits(irrad, clearsky, daily_min=0.4, daily_max=1.25):
 
     Notes
     -----
-    The default values for `daily_min` and `daily_max` were taken from
-    the PVFleets QA Analysis project.
-
     The default limits (`daily_max` and `daily_min`) have been set for
     GHI and POA irradiance for systems with *fixed* azimuth and tilt.
     If you pass POA irradiance for a tracking system it is recommended
     that you increase `daily_max` to 1.35.
+
+    The default values for `daily_min` and `daily_max` were taken from
+    the PVFleets QA Analysis project.
 
     """
     daily_irradiance = _daily_total(irrad)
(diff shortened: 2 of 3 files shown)


pvanalytics/quality/irradiance.py
```diff
@@ -390,3 +391,66 @@ def clearsky_limits(measured, clearsky, csi_max=1.1):
...
+        and `daily_max`.
+
+    Notes
+    -----
+    The default values for `daily_min` and `daily_max` were taken from
+    the PVFleets QA Analysis project.
+
+    """
+    daily_irradiance = _daily_total(irrad)
+    daily_cleasky = _daily_total(clearsky)
```
cwhanse, line 450:
```suggestion
    daily_clearsky = _daily_total(clearsky)
```

pvanalytics/quality/irradiance.py
```diff
@@ -390,3 +391,66 @@ def clearsky_limits(measured, clearsky, csi_max=1.1):
...
+    good_days = util.check_limits(
+        daily_irradiance/daily_cleasky,
```
cwhanse, line 452: The ratio is known as the clear-sky index, FYI.  Does daily_clearsky=0 raise a divide by zero warning? If so, would be good to silence it.

pvanalytics/quality/irradiance.py
```diff
@@ -390,3 +391,66 @@ def clearsky_limits(measured, clearsky, csi_max=1.1):
...
+    return pd.to_timedelta(freqstr).seconds / 3600
+
+
+def _daily_total(series):
+    freq_hours = _to_hours(pd.infer_freq(series.index))
+    return series.resample('D').sum() * freq_hours
+
+
+def daily_limits(irrad, clearsky, daily_min=0.4, daily_max=1.25):
+    """Check that daily total irradiance is near the total clearsky irradiance.
```
cwhanse, line 408: "total irradiance" is called insolation.
```suggestion
    """Check that daily insolation lies between minimum and maximum values.
```

pvanalytics/quality/irradiance.py
```diff
@@ -390,3 +391,66 @@ def clearsky_limits(measured, clearsky, csi_max=1.1):
...
+
+    Irradiance measurements on each day are integrated with the
+    left-hand rule and the total is compared to the total clearsky
+    irradiance for each day. If the ratio of total irradiance to total
+    clearsky is less than or equal to `daily_min` or greater than or
+    equal to `daily_max` then values on the day are marked False.
```
cwhanse, line 414:
```suggestion
    left-hand rule to calculate daily insolation.
```

pvanalytics/quality/irradiance.py
```diff
@@ -390,3 +391,66 @@ def clearsky_limits(measured, clearsky, csi_max=1.1):
...
+
+    .. note::
```
cwhanse, line 416: Should this go in the Notes section? Just asking

pvanalytics/quality/irradiance.py
```diff
@@ -390,3 +391,66 @@ def clearsky_limits(measured, clearsky, csi_max=1.1):
...
+    irrad : Series
+        Irradiance measurements (GHI or POA).
+    clearsky : Series
+        Expected clearsky irradiance.
+    daily_min : float, default 0.4
+        Minimum ratio of daily total irradiance to daily total
+        clearsky irradiance.
+    daily_max : float, default 1.25
+        Maximum ratio of daily total irradiance to daily total
+        clearsky irradiance.
```
cwhanse, line 434:
```suggestion
        Maximum ratio of daily insolation to daily clearsky insolation.
```

pvanalytics/quality/irradiance.py
```diff
@@ -390,3 +391,66 @@ def clearsky_limits(measured, clearsky, csi_max=1.1):
...
+
+def _daily_total(series):
+    freq_hours = _to_hours(pd.infer_freq(series.index))
+    return series.resample('D').sum() * freq_hours
+
+
+def daily_limits(irrad, clearsky, daily_min=0.4, daily_max=1.25):
+    """Check that daily total irradiance is near the total clearsky irradiance.
+
+    Irradiance measurements on each day are integrated with the
```
cwhanse, line 410:
```suggestion
    Irradiance measurements and clear-sky irradiance on each day are integrated with the
```

pvanalytics/quality/irradiance.py
```diff
@@ -390,3 +391,66 @@ def clearsky_limits(measured, clearsky, csi_max=1.1):
...
+        clearsky irradiance.
+    daily_max : float, default 1.25
+        Maximum ratio of daily total irradiance to daily total
+        clearsky irradiance.
+
+    Returns
+    -------
+    Series
+        True for values on days where the ratio of total measured
+        irradiance to total clearsky irradiance is between `daily_min`
```
cwhanse, line 440:
```suggestion
        True for values on days where the ratio of daily insolation to daily clearsky insolation is between `daily_min`
```

pvanalytics/quality/irradiance.py
```diff
@@ -390,3 +391,66 @@ def clearsky_limits(measured, clearsky, csi_max=1.1):
...
     return util.check_limits(csi, upper_bound=csi_max, inclusive_upper=True)
+
+
+def _to_hours(freqstr):
+    if freqstr[0].isalpha():
+        freqstr = '1' + freqstr
+    return pd.to_timedelta(freqstr).seconds / 3600
+
+
+def _daily_total(series):
```
cwhanse, line 402: Thinking this could be generalized in case someone would rather use a trapezoid rule. Maybe resample in the calling function and this function does the integration?

pvanalytics/quality/irradiance.py
wfvining: I think pandas will quietly return `inf` when you divide by 0, at least the version I'm working with. I'll check that the behavior is consistent though.

pvanalytics/quality/irradiance.py
wfvining: My preference would be a kwarg specifying the integration method. We can use `scipy.integration` to support the trapezoid rule and Simpson's rule as well as the left-hand rule.

pvanalytics/quality/irradiance.py
wfvining: Makes sense.

pvanalytics/quality/irradiance.py
wfvining: This is true as long as the series does not have a non-numeric dtype. I think it is safe to leave as is and allow the warning through when the Series contains strings or other wierdness.

pvanalytics/quality/irradiance.py
cwhanse: Worth trying `scipy.integration`, but it can be really slow

wfvining: One more option comes to mind. Have users pass a Series of daily insolation and provide functions in `pvanalytics.metrics` to calculate insolation either the way it is done here or with `scipy.integration`. This might give the best flexibility w.r.t. how insolation is calculated, but comes at the cost of only returning a daily mask, leaving the task of reindexing and applying the mask to the original data to the user.

cwhanse: Came across this today: https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.Series.reindex_like.html may be useful here

cwhanse: I'd suggest a trapezoid integration as the default way to get insolation from irradiance. For a regular spaced time series, it's `freq_hours * 0.5 * (series[:-1].sum() + series[1:].sum())` if that's correct code, and let's move on.

wfvining: Should we raise an error if the series does not have equal timestamp spacing (like `pvlib.clearsky.detect_clearsky`)?

cwhanse: yes, unless we want to extract the timedeltas and use them in the integral

wfvining: There may be a trade off between flexibility and performance. I would prefer to keep the flexibility of uneven timestamps, but I suspect that there could be large performance penalties for extracting the time-deltas on each day. I'll need to do a little testing to see if I'm right, or how much it matters.

wfvining: `scipy.integrate.trapz()` seems to perform very well (faster than the snippet above). I added a basic test for uniform spacing, and use the scipy function to integrate. If the timestamp spacing is uneven then the time in hours since midnight is passed to `trapz` for each value otherwise the frequency in hours is used as dx. The performance gain for having even timestamp spacing is ~4x on 8 years of 15 minute irradiance data, in both cases I think the performance is adequate.

wfvining: 84217d7 makes the performance for uneven timestamps slightly worse than what I described above; however, I am still happy since the performance of the even timestamp case is unaffected.

pvanalytics/quality/irradiance.py
```diff
@@ -390,3 +392,79 @@ def clearsky_limits(measured, clearsky, csi_max=1.1):
...
+        daily clearsky insolation is between `daily_min` and `daily_max`.
+
+    Notes
+    -----
+    The default values for `daily_min` and `daily_max` were taken from
+    the PVFleets QA Analysis project.
+
+    The default limits (`daily_max` and `daily_min`) have been set for
+    GHI. If you pass POA it is recommended that you increase
+    `daily_max` to 1.35.
```
wfvining, line 460: Re-reading the PVFleets code the daily maximum is only increased for *tracking* POA irradiance.
```suggestion
    GHI. If you pass POA irradiance for a tracking system it is
    recommended that you increase `daily_max` to 1.35.
```

wfvining, line 460: Re-reading the original PVFleets implementation the daily max is onlyt increased for POA if the system is *tracking*. For fixed POA and GHI the same limits are used.
```suggestion
    The default limits (`daily_max` and `daily_min`) have been set for
    GHI and POA irradiance for systems with *fixed* azimuth and tilt. 
    If you pass POA irradiance for a tracking system it is recommended
    that you increase `daily_max` to 1.35.
```

docs/api.rst
```diff
@@ -42,6 +42,16 @@ eliminate data that is unrealistically high.
 
    quality.irradiance.clearsky_limits
 
+You may want to identify entire days that have unrealistically high or
+low irradiance. The following function examines daily insolation,
```
cwhanse, line 46:
```suggestion
low insolation. The following function examines daily insolation,
```

pvanalytics/quality/irradiance.py
```diff
@@ -390,3 +392,80 @@ def clearsky_limits(measured, clearsky, csi_max=1.1):
...
+    return series.resample('D').apply(
+        lambda day: integrate.trapz(y=day, x=hours[day.index])
+    )
+
+
+def daily_insolation_limits(irrad, clearsky, daily_min=0.4, daily_max=1.25):
+    """Check that daily insolation lies between minimum and maximum values.
+
+    Irradiance measurements and clear-sky irradiance on each day are
+    integrated with the left-hand rule to calculate daily insolation.
```
cwhanse, line 433:
```suggestion
    integrated with the trapezoid rule to calculate daily insolation.
```

pvanalytics/quality/irradiance.py
```diff
@@ -390,3 +392,80 @@ def clearsky_limits(measured, clearsky, csi_max=1.1):
...
+
+    Parameters
+    ----------
+    irrad : Series
+        Irradiance measurements (GHI or POA).
+    clearsky : Series
+        Expected clearsky irradiance.
```
cwhanse, line 440:
```suggestion
        Clearsky irradiance.
```

pvanalytics/quality/irradiance.py
```diff
@@ -390,3 +392,80 @@ def clearsky_limits(measured, clearsky, csi_max=1.1):
...
+    daily_min : float, default 0.4
+        Minimum ratio of daily total irradiance to daily total
+        clearsky irradiance.
```
cwhanse, line 443:
```suggestion
        Minimum ratio of daily insolation to daily clearsky insolation.
```

pvanalytics/quality/irradiance.py
```diff
@@ -390,3 +392,80 @@ def clearsky_limits(measured, clearsky, csi_max=1.1):
...
+    # If the series has uniform timestamp spacing (i.e. pd.infer_freq
+    # returns a value other than None) then the frequency is used as
+    # the spacing between each value to speed up integration. If the
+    # frequency cannot be inferred, then the hour of the day is
+    # calculated for each value and used as the x-coordinate in the
+    # integration.
+    freq = pd.infer_freq(series.index)
+    if freq:
+        freq_hours = _to_hours(freq)
+        return series.resample('D').apply(
```
cwhanse, line 416: Only issue I have is not an issue but a request - help me understand how this line of code works :)

wfvining: `resample` give us one groups of all data on each day, we apply `integrate.trapz` to each group. The `dx` argument (or any other kwargs to `apply`) gets passed to `integrate.trapz` as a keyword argument.

Down below (line 424) we have to be more clever because the `x` values change for each group/application. We use a handy trick of calculating all the x values up front, capturing them in a closure, and selecting the right values for the day we are integrating.

PR #63, opened 2020-07-16
Refactor curve fitting functions

wfvining:
Since these functions are used in a `pandas.GroupBy.apply()` call, manipulating the index to calculate minute of the day can significantly reduce performance. Here we remove the `_get_minute_of_day()` function and instead pass in the x values as a parameter. In addition to improving performance this makes the curve fitting code cleaner and substantially more general.

### DIFF
--- a/pvanalytics/features/orientation.py
+++ b/pvanalytics/features/orientation.py
@@ -8,7 +8,7 @@ def _conditional_fit(day, minutes, fitfunc, freq, default=0.0, min_hours=0.0,
     # Return the :math:`r^2` of a curve fit to a single day of data if
     # certain conditions are met.
     #
-    # `fitfunc` does the curve fitting and is only applied if two 
+    # `fitfunc` does the curve fitting and is only applied if two
     # conditions are met:
     # - There must be more than `min_hours` of data in `day`
     #   (determined by the number of values in `day` times `freq`).
@@ -23,7 +23,7 @@ def _conditional_fit(day, minutes, fitfunc, freq, default=0.0, min_hours=0.0,
     # day : Series
     #     y-values to which `fitfunc` will be applied.
     # minutes : Series
-    #     x-values for curve fitting. The index for `x` must be a 
+    #     x-values for curve fitting. The index for `x` must be a
     #     superset of the index for `day`.
     # fitfunc : function
     #     Function to perform curve fit. Must accept two parameters,
--- a/pvanalytics/features/orientation.py
+++ b/pvanalytics/features/orientation.py
@@ -3,7 +3,7 @@ import pandas as pd
 from pvanalytics.util import _fit, _group
 
 
-def _conditional_fit(day, fitfunc, freq, default=0.0, min_hours=0.0,
+def _conditional_fit(day, fitfunc, minutes, freq, default=0.0, min_hours=0.0,
                      peak_min=None):
     # If there are at least `min_hours` of data in `day` and the
     # maximum for the day is greater than `peak_min` then `fitfunc` is
@@ -15,7 +15,7 @@ def _conditional_fit(day, fitfunc, freq, default=0.0, min_hours=0.0,
     if peak_min is not None:
         high_enough = day.max() > peak_min
     if (_hours(day, freq) > min_hours) and high_enough:
-        return fitfunc(day)
+        return fitfunc(minutes[day.index], day)
     return default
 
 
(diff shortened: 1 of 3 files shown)


pvanalytics/features/orientation.py
```diff
@@ -3,7 +3,7 @@
 from pvanalytics.util import _fit, _group
 
 
-def _conditional_fit(day, fitfunc, freq, default=0.0, min_hours=0.0,
+def _conditional_fit(day, fitfunc, minutes, freq, default=0.0, min_hours=0.0,
                      peak_min=None):
     # If there are at least `min_hours` of data in `day` and the
```
cwhanse, line 8: Please add docstring-like comments (or an actual docstring) to define the inputs

pvanalytics/util/_fit.py
```diff
@@ -1,26 +1,17 @@
...
-def quadratic(data):
+def quadratic(x, y):
     """Fit a quadratic to the data.
 
     Parameters
     ----------
-    data : Series
-        Series of power or irradiance measurements.
+    x : Series
+        x values for data in `y`.
```
cwhanse, line 12: Since `quadratic` and `quartic_restricted` are exposed we should provide an example of `x` that connects to the anticipated use. Is `x` a DatetimeIndex, minute of the day, or ?

wfvining: I'll add an example of `x` as minute of the day. This is currently the only way it is used.

pvanalytics/features/orientation.py
```diff
@@ -3,19 +3,51 @@
...
+    # Parameters
+    # ----------
+    # day : Series
+    #     y-values to which `fitfunc` will be applied.
+    # minutes : Series
+    #     x-values for curve fitting. Must have an index that is a
+    #     superset of the index of `day`.
+    # fitfunc : function
+    #     Function to perform curve fit. Must accept two parameters,
+    #     the x-values and y-values
```
cwhanse, line 29:
```suggestion
    #     the x-values and y-values, and return the :math:`r^2` of the curve fit.
```

pvanalytics/features/orientation.py
```diff
@@ -3,19 +3,51 @@
...
-    # If there are at least `min_hours` of data in `day` and the
-    # maximum for the day is greater than `peak_min` then `fitfunc` is
-    # applied to fit a curve to the data. `fitfunc` must be a function
-    # that takes a Series and returns the :math:`r^2` for a curve fit.
-    # If the two conditions are not met then `default` is returned and
-    # no curve fitting is performed.
+    # Return the :math:`r^2` of a curve fit to a single day of data if
+    # certain conditions are met.
+    #
+    # `fitfunc` is only applied if two conditions are met:
```
cwhanse, line 11:
```suggestion
    # `fitfunc` does the curve fitting and is only applied if two conditions are met:
```

pvanalytics/features/orientation.py
```diff
@@ -3,19 +3,51 @@
...
+    #
+    # If either condition is not satisfied then `default` is returned.
+    #
+    # Parameters
+    # ----------
+    # day : Series
+    #     y-values to which `fitfunc` will be applied.
+    # minutes : Series
+    #     x-values for curve fitting. Must have an index that is a
+    #     superset of the index of `day`.
```
cwhanse, line 26:
```suggestion
    #     x-values for curve fitting. The index for `x` must be a superset of the index for `day`.
```

pvanalytics/features/orientation.py
```diff
@@ -3,19 +3,51 @@
...
+    # fitfunc : function
+    #     Function to perform curve fit. Must accept two parameters,
+    #     the x-values and y-values
+    # freq : str
+    #     Timestamp spacing for data in `day`.
+    # default : float, default 0.0
+    #     Value to be returned if the conditions above are not
+    #     satisfied and `fitfunc` is not applied.
+    # min_hours : float, default 0.0
+    #     Minimum hours with data for curve fitting to be performed.
```
cwhanse, line 36:
```suggestion
    #     Minimum hours in `day` with data for curve fitting to be performed.
```

pvanalytics/util/_fit.py
```diff
@@ -58,8 +59,10 @@ def quartic_restricted(data, noon=720):
 
     Parameters
     ----------
-    data : Series
-        power or irradiance data.
+    x : Series
```
cwhanse, line 62:
```suggestion
    x : array or Series
```

pvanalytics/util/_fit.py
```diff
@@ -58,8 +59,10 @@ def quartic_restricted(data, noon=720):
...
+        x values for data in `y`
+    y : Series
```
cwhanse, line 64:
```suggestion
    y : array or Series
```

pvanalytics/util/_fit.py
cwhanse: ...unless you think this would require testing with array input type.

wfvining: I would feel better if we added tests with array input. Probably worth having some tests separate from the tests of orientation/tracker tests anyway.

pvanalytics/util/_fit.py
wfvining: As implemented, `y` actually needs to be a Series (we use `y.median()` for one of the restrictions).

pvanalytics/util/_fit.py
wfvining: I think the best description for this type is numpy's 'array_like'. I'll change the type for `x`, but leave `y` as a Series.

cwhanse: OK to leave as Series, since it will work with arrays but is untested.

PR #64, opened 2020-07-17
Infer azimuth and tilt from power or POA irradiance

### LINKED ISSUE #53: Determine azimuth and tilt for fixed PV systems
Given a series of power or irradiance measurements determine the azimuth and tilt of the sensor/system.

The PVFleets QA code does this by comparing the solar azimuth at the maximum of parabolas fit to the data for each day and the solar azimuth for the maximum POA irradiance of each day with different tilt and system azimuths. The tilt-azimuth pair that gives the smallest difference is returned as the tilt and azimuth of the system.

wfvining:
From scratch re-implementation of the algorithm form PVFleets QA. There are two substantial differences:
* Caller must pass a list of candidate azimuths and tilts. The PVFleets version includes an extra step where the candidate azimuths are chosen based on the data and uses a fixed set of tilts.
* The index of the solar position Series must cover the same time period as the data. In the PVFleets implementation only one year of solar position data is used. This adds some complexity that I think can be avoided by just requiring users pass in solar position for the same time period as the data. Users can pass arbitrary frequency solar position data and solar azimuth will be resampled and interpolated at 1-minute intervals before it is used. For the best results users should pass 1-minute solar position data.

closes #53

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -210,3 +210,14 @@ identification.
 
    features.orientation.fixed_nrel
    features.orientation.tracking_nrel
+
+System
+======
+
+The following function can be used to infer system orientation from
+power or plane of array irradiance measurements.
+
+.. autosummary::
+   :toctree: generated/
+
+   system.infer_orientation_least_squares
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -220,4 +220,4 @@ power or plane of array irradiance measurements.
 .. autosummary::
    :toctree: generated/
 
-   system.infer_orientation_least_squares
+   system.infer_orientation_daily_peak
--- a/pvanalytics/system.py
+++ b/pvanalytics/system.py
@@ -56,8 +56,8 @@ def infer_orientation_least_squares(power_or_poa, daytime, tilts,
     power_or_poa : Series
         Timezone localized series of power or POA irradiance
         measurements.
-    sunny : Series
-        Boolean series with True for values when it is sunny.
+    daytime : Series
+        Boolean series with True for values when it is daytime.
     tilts : list of floats
         List of candidate tilts.
     azimuths : list of floats
--- a/pvanalytics/system.py
+++ b/pvanalytics/system.py
@@ -29,6 +29,27 @@ def orientation(power_or_poa, daytime, tilts, azimuths,
                 solar_azimuth, solar_zenith, ghi, dhi, dni):
     """Determine system azimuth and tilt from power or POA.
 
+    Solar noon is estimated on each day by fitting a quadratic to data
+    in `power_or_poa` and finding the vertex of the fit. A brute force
+    search is performed on clearsky POA irradiance for all pairs of
+    candidate azimuths and tilts (`azimuths` and `tilts`) to find the
+    pair that results in the closest azimuth at solar noon to the
+    azimuths calculated from the curve fitting step. Closest is
+    determined by minimizing the sum of squared error between the
+    azimuth at solar noon on each day in `power_or_poa` and the
+    azimuth at solar noon (maximum POA) in the clearsky POA
+    irradiance.
+
+    The accuracy of the values returned from this function will vary
+    with the time-resolution of the clearsky and solar position
+    data. For the best accuracy pass `solar_azimuth`, `solar_zenith`,
+    and the clearsky data (`ghi`, `dhi`, and `dni`) with one-minute
+    timestamp spacing. If `solar_azimuth` has timestamp spacing less
+    than one minute it will be resampled and interpolated to estimate
+    azimuth at each minute of the day. Regardless of the timestamp
+    spacing these parameters must cover the same days as
+    `power_or_poa`.
+
     Parameters
     ----------
     power_or_poa : Series
@@ -37,9 +58,9 @@ def orientation(power_or_poa, daytime, tilts, azimuths,
     sunny : Series
         Boolean series with True for values when it is sunny.
     tilts : list of floats
-        list of tilts to check
+        List of candidate tilts.
     azimuths : list of floats
-        list of azimuths
+        List of candidate azimuths.
     solar_azimuth : Series
         Time series of solar azimuth.
     solar_zenith : Series
@@ -55,6 +76,10 @@ def orientation(power_or_poa, daytime, tilts, azimuths,
     -------
     azimuth : float
     tilt : float
+
+    Notes
+    -----
+    Based on PVFleets QA project.
 
     """
     peak_times = _peak_times(power_or_poa[daytime])
--- a/pvanalytics/system.py
+++ b/pvanalytics/system.py
@@ -77,7 +77,7 @@ def orientation(power_or_poa, daytime, tilts, azimuths,
                 dni=dni
             ).poa_global
             poa_azimuths = azimuth_by_minute[
-                _peak_times(poa[solar_zenith < 70])
+                _group.by_day(poa).idxmax()
             ]
             filtered_azimuths = poa_azimuths[np.isin(
                 poa_azimuths.index.date,
--- a/pvanalytics/system.py
+++ b/pvanalytics/system.py
@@ -28,7 +28,7 @@ def _peak_times(data):
 def infer_orientation_solarnoon(power_or_poa, daytime, tilts,
                                 azimuths, solar_azimuth,
                                 solar_zenith, ghi, dhi, dni):
-    """Determine system azimuth and tilt from power or POA using apparent
+    """Determine system azimuth and tilt from power or POA using inferred
     solar noon.
 
     Solar noon is estimated on each day by fitting a quadratic to data
--- a/pvanalytics/system.py
+++ b/pvanalytics/system.py
@@ -35,9 +35,9 @@ def infer_orientation_solarnoon(power_or_poa, daytime, tilts,
     in `power_or_poa` and finding the vertex of the fit. A brute force
     search is performed on clearsky POA irradiance for all pairs of
     candidate azimuths and tilts (`azimuths` and `tilts`) to find the
-    pair that results in the closest azimuth at solar noon to the
-    azimuths calculated from the curve fitting step. Closest is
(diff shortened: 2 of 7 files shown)


wfvining:
Converting this to a draft. It needs more testing, there seems to be a bug where the `modeled_azimuths` ends up with a different length than `poa_azimuths`.

wfvining:
I thought this bug might be caused by gaps in the data, but it looks like that's not it. It may be user error: a slight mismatch in the indexes of the data and the clearsky/solarposition data. Even if that is the case we should detect it and either raise an exception with a useful error message (current error message: `ValueError: operands could not be broadcast together with shapes (xxx,) (yyy,)`) or find a way to handle the mismatch.

wfvining:
The bug is caused by having gaps in the measurements, but not the clearsky or solarposition data. A quick fix is to filter the gaps out data generated for the candidate azimuths (line 83 ish):

```python
            sum_of_squares = sum(
                (poa_azimuths[modeled_azimuth.index].values
                 - modeled_azimuth.values)**2
            )
```

pvanalytics/util/_fit.py
```diff
@@ -3,6 +3,36 @@
...
+    """Fit a quartic to the data and return the x-value of the vertex.
+
+    Parameters
+    ----------
+    x : array_like
+    y : Series
+        x and y-values to fit to.
+    model_range : array_like, optional
+        Range of values to find the max in. If not specified then `x`
+        is used.
```
wfvining, line 22:
```suggestion
        Range of x-values to find the max over. If not specified then
        `x` is used.
```

pvanalytics/util/_fit.py
```diff
@@ -3,6 +3,36 @@
 import scipy.optimize
 
 
+def _quadratic(xs, ys):
+    # fit a quadratic function of `xs` to the data in `ys`
+    coefficients = np.polyfit(xs, ys, 2)
+    return np.poly1d(coefficients)
+
+
+def quadratic_idxmax(x, y, model_range=None):
```
wfvining, line 12: Add an example to the docstring.

pvanalytics/system.py
```diff
@@ -1 +1,118 @@
...
+            poa_azimuths = azimuth_by_minute[
+                _group.by_day(poa).idxmax()
+            ]
+            filtered_azimuths = poa_azimuths[np.isin(
+                poa_azimuths.index.date,
+                modeled_azimuth.index.date
+            )]
+            sum_of_squares = sum(
+                (filtered_azimuths.values - modeled_azimuth.values)**2
+            )
```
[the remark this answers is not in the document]
wfvining: This won't actually work since the `modeled_azimuth` series includes the same dates, but potentially different times from `poa_azimuths`. After trying a few things out I think it is best to leave it as is.

wfvining:
The only component of the PVFleets algorithm that is not implemented is automatically generating a list of candidate azimuths. Because PVFleets is focused on North America, the method is obviously specific to the Northern Hemisphere. I haven't been able to find a satisfactory way to make the method generic enough to fit in this library.

pvanalytics/util/_fit.py
```diff
@@ -3,6 +3,49 @@
...
+    ... )
+    0
+
+    """
+    model_range = model_range or x
+    quadratic = _quadratic(x, y)
+    model = quadratic(model_range)
+    return model_range[np.argmax(model)]
+
+
```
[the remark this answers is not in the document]
wfvining: Yes, I agree

wfvining:
@cwhanse `model_range` might be a bad name. It should probably be called `model_domain` or `model_xs`. I added it as a parameter so the curve fitting functions could be very general and find the maximum over an arbitrary range of x-values. What are you concerns with the parameter?

pvanalytics/system.py
```diff
@@ -1 +1,119 @@
...
+    """Determine system azimuth and tilt from power or POA using least
+    squares.
+
+    Solar noon is estimated on each day by fitting a quadratic to data
+    in `power_or_poa` and finding the vertex of the fit. A brute force
+    search is performed on clearsky POA irradiance for all pairs of
+    candidate azimuths and tilts (`azimuths` and `tilts`) to find the
+    pair that results in the closest azimuth at solar noon to the
+    azimuths calculated from the curve fitting step. Closest is
+    determined by minimizing the sum of squared error between the
```
cwhanse, line 40:
```suggestion
    determined by minimizing the sum of squared difference between the
```

pvanalytics/system.py
```diff
@@ -1 +1,119 @@
...
+                                    solar_zenith, ghi, dhi, dni):
+    """Determine system azimuth and tilt from power or POA using least
+    squares.
+
+    Solar noon is estimated on each day by fitting a quadratic to data
+    in `power_or_poa` and finding the vertex of the fit. A brute force
+    search is performed on clearsky POA irradiance for all pairs of
+    candidate azimuths and tilts (`azimuths` and `tilts`) to find the
+    pair that results in the closest azimuth at solar noon to the
+    azimuths calculated from the curve fitting step. Closest is
```
cwhanse, line 39:
```suggestion
    pair that results in the closest azimuth to the
    azimuths calculated at solar noon  from the curve fitting step. Closest is
```

pvanalytics/system.py
```diff
@@ -1 +1,119 @@
...
+    spacing these parameters must cover the same days as
+    `power_or_poa`.
+
+    Parameters
+    ----------
+    power_or_poa : Series
+        Timezone localized series of power or POA irradiance
+        measurements.
+    daytime : Series
+        Boolean series with True for values when it is daytime.
```
cwhanse, line 60: Could this be inferred from `solar_zenith` with a zenith limit? Maybe that has been tried.

pvanalytics/system.py
```diff
@@ -1 +1,119 @@
...
+            )
+        )
+    )
+    return pd.DatetimeIndex(
+        np.unique(data.index.date),
+        tz=data.index.tz
+    ) + peak_minutes
+
+
+def infer_orientation_least_squares(power_or_poa, daytime, tilts,
```
cwhanse, line 28: I think I would prefer `_solarnoon` or `_pvfleets`, `_least_squares` is generic and could be a feature of other tilt/azimuth algorithms.

cwhanse:
> @cwhanse `model_range` might be a bad name. It should probably be called `model_domain` or `model_xs`. I added it as a parameter so the curve fitting functions could be very general and find the maximum over an arbitrary range of x-values. 

`model_domain` or just `domain` would be better,.

> What are you concerns with the parameter?

The description of `quadratic_idmax` says that the function returns the vertex. It doesn't, it returns the value in `model_range` at which the fitted quadratic is maximized. If only morning data were passed, the quadratic's vertex is to the right of the data (assuming the data are concave down). If `model_domain` is only morning data, the function would return the right-most value from `model_range`. If `quadratic_idmax` is supposed to return the value in `domain` where the fit is maximized, OK, just adjust the description.

Alternatively, if we want `quadratic_idmax` to return the vertex of the fit, it should return -b/2a and then let the calling function decide what to do with that value.

pvanalytics/system.py
wfvining: I was a little confused by this parameter after your last review so I went back to my notes. It was originally called 'sunny' because it was supposed to mask out cloudy days as well as night. I may have gotten a little overzealous in my renaming of parameters from the original PVFleets implementation. Should probably go back and rename this param `clearsky` or something along those lines.

cwhanse: Aha, this algorithm makes a lot more sense if `power_or_poa` has already been filtered to times with clear-sky conditions, or `sunny` is a filter that subsets `power_or_poa` to clear-sky conditions.

wfvining: I'll change the parameter name back to `sunny` or `clearsky` and write a better description.

pvanalytics/system.py
```diff
@@ -25,11 +25,11 @@ def _peak_times(data):
...
 
-def infer_orientation_least_squares(power_or_poa, daytime, tilts,
-                                    azimuths, solar_azimuth,
-                                    solar_zenith, ghi, dhi, dni):
-    """Determine system azimuth and tilt from power or POA using least
-    squares.
+def infer_orientation_solarnoon(power_or_poa, daytime, tilts,
+                                azimuths, solar_azimuth,
+                                solar_zenith, ghi, dhi, dni):
+    """Determine system azimuth and tilt from power or POA using apparent
```
cwhanse, line 31:
```suggestion
    """Determine system azimuth and tilt from power or POA using inferred
```

'apparent' has a specific meaning in solar position.

wfvining:
> Alternatively, if we want quadratic_idmax to return the vertex of the fit, it should return -b/2a and then let the calling function decide what to do with that value.

I think this is the right thing to do in the interest of keeping the code in `_fit` as general as possible. The name `quadratic_idxmax()` doesn't really make sense with that change so I think I will rename it to `quadratic_vertex()`.

wfvining:
Possible names: `infer_orientation_poanoon` or `infer_orientation_poamax`

pvanalytics/system.py
```diff
@@ -1 +1,121 @@
...
+            )
+        )
+    )
+    return pd.DatetimeIndex(
+        np.unique(data.index.date),
+        tz=data.index.tz
+    ) + peak_minutes
+
+
+def infer_orientation_solarnoon(power_or_poa, sunny, tilts,
```
cwhanse, line 29: how about `_daily_peak`?  If you agree, the `Solar noon is estimated` -> `The time of the daily peak is estimated`

docs/api.rst
```diff
@@ -210,3 +210,14 @@ identification.
...
+System
+======
+
+The following function can be used to infer system orientation from
+power or plane of array irradiance measurements.
+
+.. autosummary::
+   :toctree: generated/
+
+   system.infer_orientation_least_squares
```
cwhanse, line 223: update to match final name

PR #67, opened 2020-08-11
Identify when the sun is up based on power or irradiance

wfvining:
## Description

Flag daytime periods based on power or irradiance data. Requires no knowledge of location or timezone. Reimplimentation of algorithm from PVFleets with some modifications. In a nutshell it looks for times where the derivative (approximated as a first-order difference) is non-zero and the data is large. Can handle intermittent periods of low values/low slope during the day by looking for too-short daytime periods and comparing to the day/night mask value on preceding and following days. Can be combined with a clipping mask to ensure clipping is flagged as daytime.

## Checklist

- [x] consolidate fixtures (there is a fair amount of duplication in the new tests that should be eliminated)
- [x] add test with outlier mask
- [x] test with some variable days
- [x] Either limit pandas version to < 1.1  or wait for 1.1.1 to be released ([bug in 1.1.0](https://github.com/pandas-dev/pandas/issues/35552))

### DIFF
--- a/pvanalytics/features/daytime.py
+++ b/pvanalytics/features/daytime.py
@@ -1,0 +1,5 @@
+"""Functions for identifying daytime"""
+
+
+def diff(data):
+    return data > 0
--- a/pvanalytics/features/daytime.py
+++ b/pvanalytics/features/daytime.py
@@ -61,20 +61,49 @@ def _from_numeric(series):
     return boolean_series.astype('bool')
 
 
-def _filter_midday_errors(night, minutes_per_value):
-    # identify periods of time that appear to switch from night to day
-    # (or day to night) on too short a time scale to be reasonable.
-    invalid = _run_lengths(night)*minutes_per_value <= 5*60  # 5 hours
-    # Need a numeric series so we can use Series.median()
-    numeric_series = _to_numeric(night)
+def _smooth_if_invalid(series, invalid):
+    # replace values in `series` marked as True in `invalid` with the
+    # 31-day rolling median at each minute.
     smoothed = round(
         _rolling_by_minute(
-            numeric_series,
+            _to_numeric(series),
             days=31,
             f=pd.core.window.RollingGroupby.median
         )
     )
-    return (~invalid & night) | (invalid & _from_numeric(smoothed))
+    return (~invalid & series) | (invalid & _from_numeric(smoothed))
+
+
+def _filter_midday_errors(night, minutes_per_value):
+    # identify periods of time that appear to switch from night to day
+    # (or day to night) on too short a time scale to be reasonable.
+    invalid = _run_lengths(night)*minutes_per_value <= 5*60  # 5 hours
+    return _smooth_if_invalid(night, invalid)
+
+
+def _filter_edge_of_day_errors(night, minutes_per_value):
+    # identify night-time periods that are "too long" and replace
+    # values with the 31-day rolling median value for that minute.
+    #
+    # Because daylight savings shifts happen at night we cannot look
+    # at night-length directly. Instead we look for too-short days and
+    # flag the full day for substitution. This may result in slightly
+    # reduced accuracy for sunrise/sunset times on these days (even if
+    # one end of the day - sunrise or sunset - was correctly flagged,
+    # it will be replaced with the rolling median for that minute).
+    day_length = night.groupby(night.cumsum()).transform(
+        lambda x: len(x) * minutes_per_value
+    )
+    # remove night time values so they don't interfer with the mediand
+    # day length.
+    day_length.loc[night] = np.nan
+    day_length_median = day_length.rolling(window='14D').median()
+    # flag days that are more than 30 minutes shorter than the median
+    short_days = day_length < (day_length_median - 30)
+    invalid = short_days.groupby(short_days.index.date).transform(
+        lambda day: any(day)
+    )
+    return _smooth_if_invalid(night, invalid)
 
 
 def _filter_and_normalize(series, outliers):
@@ -142,9 +171,12 @@ def diff(series, outliers=None, low_value_threshold=0.003,
              | (low_value & low_median)
              | (low_diff & low_median))
     # Fix erroneous classifications (e.g. midday outages where power
-    # goes to 0 and stays there for several hours)
-    night = _filter_midday_errors(night, minutes_per_value)
-    # TODO optional validation against clearsky daylight hours
-    # If a clipping mask was provided mark all clipped values as
-    # daytime
-    return (clipping or False) | (~night)
+    # goes to 0 and stays there for several hours, clipping classified
+    # as night, and night-time periods that are too long)
+    night_corrected_midday = _filter_midday_errors(night, minutes_per_value)
+    night_corrected_clipping = ~((clipping or False)
+                                 | (~night_corrected_midday))
+    night_corrected_edges = _filter_edge_of_day_errors(
+        night_corrected_clipping, minutes_per_value
+    )
+    return ~night_corrected_edges
--- a/pvanalytics/features/daytime.py
+++ b/pvanalytics/features/daytime.py
@@ -109,7 +109,8 @@ def _filter_edge_of_day_errors(night, minutes_per_value):
 def _filter_and_normalize(series, outliers):
     # filter a series by removing outliers and clamping the minimum to
     # 0. Then normalize the series by the maximum deviation.
-    series = series[~outliers]
+    if outliers is not None:
+        series.loc[outliers] = np.nan
     series.loc[series < 0] = 0
     return (series - series.min()) / (series.max() - series.min())
 
@@ -154,10 +155,7 @@ def diff(series, outliers=None, low_value_threshold=0.003,
     Derived from the PVFleets QA Analysis project.
     """
     series = series.fillna(value=0)
-    series_norm = _filter_and_normalize(
-        series,
-        outliers or pd.Series(False, index=series.index)
-    )
+    series_norm = _filter_and_normalize(series, outliers).fillna(value=0)
     minutes_per_value = _freqstr_to_minutes(
         freq or pd.infer_freq(series.index)
     )
--- a/pvanalytics/features/daytime.py
+++ b/pvanalytics/features/daytime.py
@@ -63,7 +63,7 @@ def _correct_midday_errors(night, minutes_per_value, hours_min,
 
 
 def _correct_edge_of_day_errors(night, minutes_per_value,
-                                daytime_difference_max,
+                                day_length_difference_max,
                                 day_length_window, correction_window):
     # Identify day-time periods that are "too short" and replace
     # values with the majority truth-value for the same time in the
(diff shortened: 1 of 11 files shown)


pvanalytics/features/daytime.py
```diff
@@ -0,0 +1,210 @@
...
+    ----------
+    series : Series
+        Time series of power or irradiance.
+    outliers : Series, optional
+        Boolean time series with True for values in `series` that are
+        outliers.
+    low_value_threshold : float, default 0.003
+        Maximum normalized value for a time to be considered night.
+    low_median_threshold : float, default 0.0015
+        Minimum rolling median for a time to be considered night.
```
wfvining, line 157: The rolling median should be low at night. This should be a maximum threshold
```suggestion
        Maximum rolling median for a time to be considered night.
```

pvanalytics/features/daytime.py
```diff
@@ -0,0 +1,210 @@
...
+             | (low_diff & low_median))
+    # Fix erroneous classifications (e.g. midday outages where power
+    # goes to 0 and stays there for several hours, clipping classified
+    # as night, and night-time periods that are too long)
+    night_corrected_midday = _filter_midday_errors(night, minutes_per_value)
+    night_corrected_clipping = ~((clipping or False)
+                                 | (~night_corrected_midday))
+    night_corrected_edges = _filter_edge_of_day_errors(
+        night_corrected_clipping, minutes_per_value
+    )
```
wfvining, line 209: A few thoughts:

- Rename these functions "filter" isn't the right word. Let's call them `_correct_*_errors()`.
- Add a comment describing why we call `_filter_edge_of_day_errors()` and maybe also why that function is separate from `_filter_midday_errors()`
- Should the clipping correction come first? It seems like that would give more information to the two error correction algorithms.

wfvining: Is there any value in making any of the correction steps optional? (e.g. pass Booleans `correct_midday`, and `correct_edge_of_day`)

pvanalytics/tests/features/test_daytime.py
```diff
@@ -0,0 +1,187 @@
...
+import pytest
+import pandas as pd
+import numpy as np
+from pvlib.location import Location
+from pvanalytics.features import daytime
+
+
+@pytest.fixture(scope='module')
+def albuquerque():
+    return Location(35, -106, altitude=1500)
```
wfvining, line 11: Super generic and used elsewhere. Should maybe be moved to `conftest.py`

wfvining:
Discussed possibility of using an optional clearsky or lat/lon parameter for correcting morning & evening errors with @matt14muller. Using the rolling median/mean of the day lengths can exacerbate problems when there are intermittent errors in the data that skew the mean/median. If you have even approximate knowledge of the location of the system clearsky or solar transit time data can give you a much better result in these cases.

The original PVFleets algorithm uses the expected number of hours of daylight from pvlib to identify problem days rather than comparing to a rolling mean. The reason for not doing that here was to attempt to provide an algorithm that will work without knowledge of *where* the system is.

pvanalytics/features/daytime.py
```diff
@@ -0,0 +1,247 @@
...
+def _from_numeric(series):
+    # Inverse of :py:func:`_to_numeric()`
+    boolean_series = series == 2
+    boolean_series.loc[series == 3] = np.nan
+    return boolean_series.astype('bool')
+
+
+def _smooth_if_invalid(series, invalid, correction_window):
+    # replace values in `series` marked as True in `invalid` with the
+    # `correction_window`-day rolling median at each minute.
```
cwhanse, line 64: seems like a different term than 'median' would help, to avoid confusing this step with the 'low median' criteria. What is being done (if I follow correctly) is a voting algorithm, where a minute in an invalid window is relabeled to be the same as the majority label of minutes in the correction_window

pvanalytics/features/daytime.py
```diff
@@ -0,0 +1,247 @@
...
+        Maximum derivative for a time to be considered night.
+    median_days : int, default 7
+        Number of days to use to calculate the rolling median at each
+        minute.
+    clipping : Series, optional
+        True when clipping indicated. Any values where clipping is
+        indicated are automatically considered 'daytime'.
+    correction_window : int, default 31
+        Number of adjacent days to examine when correcting
+        day/night classification errors.
```
cwhanse, line 180: May help to note that this should be an odd value

pvanalytics/features/daytime.py
```diff
@@ -0,0 +1,247 @@
...
+    invalid = _run_lengths(night)*minutes_per_value <= hours_min*60
+    return _smooth_if_invalid(night, invalid, correction_window)
+
+
+def _correct_edge_of_day_errors(night, minutes_per_value,
+                                daytime_difference_max,
+                                day_length_window, correction_window):
+    # identify night-time periods that are "too long" and replace
+    # values with the `correction_window`-day rolling median value for
+    # that minute.
```
cwhanse, line 88:
```suggestion
    # identify day-time periods that are "too short" and replace
    # values with the `correction_window`-day rolling median value for
    # that minute.
```

pvanalytics/features/daytime.py
```diff
@@ -0,0 +1,247 @@
...
+    clipping : Series, optional
+        True when clipping indicated. Any values where clipping is
+        indicated are automatically considered 'daytime'.
+    correction_window : int, default 31
+        Number of adjacent days to examine when correcting
+        day/night classification errors.
+    hours_min : float, default 5
+        Minimum number of hours in a contiguous period of day or
+        night. If a day/night period is shorter than this then it is
+        flagged for error correction.
```
cwhanse, line 184:
```suggestion
        flagged for error correction. [hours]
```

pvanalytics/features/daytime.py
cwhanse, line 180:
```suggestion
        day/night classification errors. [days]
```

pvanalytics/features/daytime.py
```diff
@@ -0,0 +1,247 @@
...
+        outliers.
+    low_value_threshold : float, default 0.003
+        Maximum normalized value for a time to be considered night.
+    low_median_threshold : float, default 0.0015
+        Maximum rolling median for a time to be considered night.
+    low_diff_threshold : float, default 0.0005
+        Maximum derivative for a time to be considered night.
+    median_days : int, default 7
+        Number of days to use to calculate the rolling median at each
+        minute.
```
cwhanse, line 174:
```suggestion
        minute. [days]
```

pvanalytics/features/daytime.py
```diff
@@ -0,0 +1,247 @@
...
+        Time series of power or irradiance.
+    outliers : Series, optional
+        Boolean time series with True for values in `series` that are
+        outliers.
+    low_value_threshold : float, default 0.003
+        Maximum normalized value for a time to be considered night.
+    low_median_threshold : float, default 0.0015
+        Maximum rolling median for a time to be considered night.
+    low_diff_threshold : float, default 0.0005
+        Maximum derivative for a time to be considered night.
```
cwhanse, line 171:
```suggestion
        Maximum normalized power or irradiance value for a time to be considered night.
    low_median_threshold : float, default 0.0015
        Maximum rolling median of power or irradiance for a time to be considered night.
    low_diff_threshold : float, default 0.0005
        Maximum derivative of normalized power or irradiance for a time to be considered night.
```

pvanalytics/features/daytime.py
```diff
@@ -0,0 +1,247 @@
...
+        True when clipping indicated. Any values where clipping is
+        indicated are automatically considered 'daytime'.
+    correction_window : int, default 31
+        Number of adjacent days to examine when correcting
+        day/night classification errors.
+    hours_min : float, default 5
+        Minimum number of hours in a contiguous period of day or
+        night. If a day/night period is shorter than this then it is
+        flagged for error correction.
+    daytime_difference_max : float, default 30
```
cwhanse, line 185: Maybe 'day_length_difference_max'?

pvanalytics/features/daytime.py
```diff
@@ -0,0 +1,247 @@
...
+        When looking for day that end too early or start too late any
+        day where the hours of daylight is more then
+        `daytime_difference_max` minutes shorter than the median
+        length of surrounding days.
```
cwhanse, line 189:
```suggestion
        Days with length that is `daytime_difference_max` minutes less than the median
        length of surrounding days are flagged for corrections.
```

pvanalytics/features/daytime.py
```diff
@@ -0,0 +1,247 @@
...
+    day_length_window : int, default 14
+        The length of the rolling window used for calculating the
+        median length of the day when correcting errors in the morning
+        or afternoon.
```
cwhanse, line 193:
```suggestion
        or afternoon. [days]
```

pvanalytics/features/daytime.py
wfvining: I don't think that it has to be odd. If the window is even, the number of values preceding the label will just be one more than the number of following values.

pvanalytics/features/daytime.py
```diff
@@ -0,0 +1,247 @@
...
+    # (or day to night) on too short a time scale to be reasonable.
+    invalid = _run_lengths(night)*minutes_per_value <= hours_min*60
+    return _smooth_if_invalid(night, invalid, correction_window)
+
+
+def _correct_edge_of_day_errors(night, minutes_per_value,
+                                daytime_difference_max,
+                                day_length_window, correction_window):
+    # identify night-time periods that are "too long" and replace
+    # values with the `correction_window`-day rolling median value for
```
wfvining, line 87: See note in `_correct_midday_errors()` regarding "median"

pvanalytics/features/daytime.py
```diff
@@ -0,0 +1,247 @@
...
+    # - Near-zero value
+    # - Near-zero first-order derivative
+    # - Near-zero rolling median
+    low_value = series_norm <= low_value_threshold
+    low_median = rolling_median <= low_median_threshold
+    low_diff = abs(first_order_diff) <= low_diff_threshold
+
+    night = ((low_value & low_diff)
+             | (low_value & low_median)
+             | (low_diff & low_median))
```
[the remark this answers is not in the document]
wfvining: Applying `Series.sum()` to a Boolean series will return a numeric value, but if you use `+` on two Boolean series pandas will interpret it as a logical *OR* and return a Boolean series. To make this work we would have to convert the masks to numeric values first which I think would negate any readability benefits.

requirements.txt
```diff
@@ -1,4 +1,4 @@
 numpy>=1.16.0
-pandas>=0.25.0
+pandas>=0.25.0,<1.1.0
```
wfvining, line 2: 1.1.1 is out. Should see if we can relax this constraint.

wfvining: Upgrading to 1.1.1 results in test failures for `test_daytime_split_day` and `test_daytime_daylight_savings`. The reason for the failures is not immediately clear.

wfvining: `rolling().groupby(center=True)` still broken in pandas > 1.0. https://github.com/pandas-dev/pandas/issues/36040

cwhanse (approved):
Looks good

PR #74, opened 2020-09-21
Time shift detection

wfvining:
## Description

Look for time-shifts (such as daylight savings) by comparing midday times derived from a boolean day/night mask and expected midday times (e.g. calculated form a solar position algorithm).

## Checklist

- [x] consider adding a test where the day/night mask is not tz-localized.


wfvining:
The coverage decrease is expected. There is a conditional that is never satisfied, but I added due to a lack of clarity in the ruptures documentation. The test is to ensure that the end of the series is always identified as a break-point so that the intervals in which shifts are calculated are always closed on both ends.

It may be reasonable to remove this conditional since I truly have not found a way to get ruptures to return a value where it would be satisfied.

requirements.txt
```diff
@@ -2,3 +2,4 @@ numpy>=1.16.0
 pandas>=0.25.0
 pvlib>=0.7.0
 scipy>=1.3.0
+ruptures
```
cwhanse, line 5: May want to make this an optional dependency, check for import in the `shifts_ruptures` function and raise an error if not found.

wfvining: I think that is a good idea.

pvanalytics/quality/time.py
```diff
@@ -33,3 +36,130 @@ def spacing(times, freq):
...
+    `shift_min` and returned as the time-shift for all values in that
+    period.
+
+    Parameters
+    ----------
+    daytime : Series
+        Boolean time-series with True for day and False for night.
+    clearsky_midday : Series
+        Time of midday in minutes for each day with no time shifts
+        (i.e.based on solar position for with a fixed-offset time zone).
```
wfvining, line 83:
```suggestion
        (i.e.based on solar position with a fixed-offset time zone).
```

pvanalytics/quality/time.py
```diff
@@ -33,3 +36,130 @@ def spacing(times, freq):
...
+
+    midday = _group.by_day(daytime).apply(
+        lambda day: (day[day].index.min()
+                     + ((day[day].index.max() - day[day].index.min()) / 2))
+    )
+    if period_min > len(midday):
+        raise ValueError("period_min exceeds number of days in series")
+    midday_minutes = midday.dt.hour * 60 + midday.dt.minute
+    # Drop timezone information. At this point there is one value per day
+    # so it is no longer needed.
```
wfvining, line 134:
```suggestion
    # so the timezone is irrelevant.
```

pvanalytics/quality/time.py
```diff
@@ -33,3 +36,130 @@ def spacing(times, freq):
...
+    midday_diff = \
+        midday_minutes.tz_localize(None) - clearsky_midday.tz_localize(None)
+    break_points = ruptures.Pelt(
+        model='rbf',
+        jump=1,
```
wfvining, line 139: Do we want to make these parameters?

pvanalytics/quality/time.py
```diff
@@ -33,3 +36,130 @@ def spacing(times, freq):
...
+        pd.cut(
+            midday_diff.reset_index().index,
+            break_points,
+            include_lowest=True, right=False,
+            duplicates='drop'
+        )
+    ).transform(
+        lambda shifted_period: stats.mode(shifted_period).mode[0]
+    )
+    # localize the shit amount series to the timezone of the input
```
wfvining, line 163:
```suggestion
    # localize the shift_amount series to the timezone of the input
```

pvanalytics/quality/time.py
```diff
@@ -33,3 +36,130 @@ def spacing(times, freq):
...
+
+    """
+    try:
+        import ruptures
+    except ImportError:
+        raise ImportError("time.shifts_ruptures() requires ruptures.")
+
+    midday = _group.by_day(daytime).apply(
+        lambda day: (day[day].index.min()
+                     + ((day[day].index.max() - day[day].index.min()) / 2))
```
cwhanse, line 128: Not a request to change, but I don't follow `day[day]` here - isn't `day` itself a slice from `daytime`?

wfvining: This is just a simple way to get the first time-stamp that has a `True` value. Since `day` is a Boolean series we can select a slice using itself to get only the daytime timestamps then take the minimum.

Maybe it would be easier to read if we remove the night-time timestamps before the `groupby` operation. Would look something like:

```python
    _group.by_day(daytime[daytime]).apply(lambda day: day.index.min() + (day.index.max() - day.index.min()) / 2)
```

pvanalytics/quality/time.py
cwhanse: Ok not to at this stage

pvanalytics/quality/time.py
```diff
@@ -33,3 +36,130 @@ def spacing(times, freq):
...
+    midday times from `clearsky_midday`. For each period between change
+    points the mode of the difference is rounded to a multiple of
+    `shift_min` and returned as the time-shift for all values in that
+    period.
+
+    Parameters
+    ----------
+    daytime : Series
+        Boolean time-series with True for day and False for night.
+    clearsky_midday : Series
```
cwhanse, line 81: Thinking if this input would be better named `transit` for the solar transit time, or local noon. It doesn't have much to do with clearsky conditions, it's just the midpoint between sunrise and sunset.

pvanalytics/quality/time.py
```diff
@@ -33,3 +36,130 @@ def spacing(times, freq):
...
+    The Pelt changepoint detection method is applied to the difference
+    in midday times derived from the `daytime` series and the expected
+    midday times from `clearsky_midday`. For each period between change
+    points the mode of the difference is rounded to a multiple of
+    `shift_min` and returned as the time-shift for all values in that
+    period.
+
+    Parameters
+    ----------
+    daytime : Series
```
cwhanse, line 79: If I get this right, the function detects change points in a time series derived from `daytime`, i.e., the midpoints between first True and first False for each day. Using `daytime` as the primary input may be the PVFleets usage, but it seems awkward to reduce a time series of irradiance/power to this daytime boolean to be able to use this function. I'm wondering if the primary input could be just the midday minutes.

pvanalytics/quality/time.py
```diff
@@ -33,3 +36,130 @@ def spacing(times, freq):
...
+        Boolean time-series with True for day and False for night.
+    clearsky_midday : Series
+        Time of midday in minutes for each day with no time shifts
+        (i.e.based on solar position with a fixed-offset time zone).
+    period_min : int, default 2
+        Minimum number of days between shifts. Must be less than or equal to
+        the number of days in `daytime`.
```
cwhanse, line 86:
```suggestion
        the number of days in `daytime`. [days]
```

pvanalytics/quality/time.py
```diff
@@ -33,3 +36,130 @@ def spacing(times, freq):
...
+    shift_min : int, default 15
+        Minimum shift amount in minutes. All shifts are rounded to a multiple
+        of `shift_min`. [minutes]
+    round_up_from : int, optional
+        The number of minutes greater than a multiple of `shift_min` for a
+        shift to be rounded up. If a shift is less than `round_up_from` then
+        it will be rounded towards 0. If not specified then the shift will
+        be rounded up from `shift_min // 2`. Using a larger value will
+        effectively make the shift detection more conservative as small
+        variations will tend to be rounded to zero.
```
cwhanse, line 101:
```suggestion
        variations will tend to be rounded to zero. [minutes]
```

pvanalytics/quality/time.py
```diff
@@ -33,3 +36,130 @@ def spacing(times, freq):
...
+
+    Raises
+    ------
+    ValueError
+        If the number of days in `daytime` is less than `period_min`.
+
+    Notes
+    -----
+    Derived from the PVFleets QA project.
+
```
cwhanse, line 119: This docstring would really benefit from an example, or, a paragraph describing a possible use of this function.

Tossing some text out:

Timestamped data from monitored PV systems may not always be localized to a consistent timezone. In some cases, data is timestamped with local time that may or may not be adjusted for daylight savings time transitions.  This function helps detect issues of this sort, by detecting points where the difference in times at a consistent point (solar noon) changes significantly. If the data's timestamps have not been adjusted for daylight savings transitions, the time of day at solar noon will change by roughly 60 minutes in the days before and after the transition.

To use this changepoint detection method to determine if your data's timestamps involve daylight savings transitions, first reduce your PV system data (irradiance or power) to a daily time series, with each point being the observed midday time in minutes. For example, if sunrise and sunset are inferred from the PV system data, the midday time can be inferred as the average of each day's sunrise and sunset time of day. To establish the expected midday time, calculate solar transit time in time of day. This function detects shifts in the difference between the observed and expected midday times, and returns  **(here I'm unclear what is being returned)**

cwhanse (commented):
Progress. We may want to adjust the inputs, and I'm not sure what is contained in the returned Series. It is not a list of rupture points, or Boolean if there's evidence of a timezone/daylight savings transition.

pvanalytics/quality/time.py
wfvining: I think your right. requiring a Boolean series is probably less flexible than midday minutes. We may want to provide a utility function in `features.daytime` that will convert the boolean series to midday minutes.

pvanalytics/quality/time.py
```diff
@@ -33,3 +36,130 @@ def spacing(times, freq):
...
+    `shift_min` and returned as the time-shift for all values in that
+    period.
+
+    Parameters
+    ----------
+    daytime : Series
+        Boolean time-series with True for day and False for night.
+    clearsky_midday : Series
+        Time of midday in minutes for each day with no time shifts
+        (i.e.based on solar position with a fixed-offset time zone).
```
wfvining, line 83:
```suggestion
        (i.e. based on solar position with a fixed-offset time zone).
```

pvanalytics/quality/time.py
wfvining: The return value is a series of minutes that if added to the timestamps of the series will effectively put the entire series into the same time zone as the midday times in `clearsky_midday` (to be renamed `transit`).

Currently returns a series of integers. Perhaps it would be better to return a series of `Timedelta`.

cwhanse: I guess this speaks to the function's anticipated use: to detect timezone issues, or to correct them? Returning a series of integers or `Timedelta` is a correction, one would need to inspect the values to know if there is an issue. Maybe return both?  A Series of bool True for timestamps that are out of synch with expected, and a Series of suggested corrections. I'm not yet convinced that the corrections are desirable, since they rely on the observed transit values.

wfvining: Returning both is probably the most useful thing at this point. The issue with returning a boolean series, however, is that if I pass `transit` in central time, for example, and `midday` in Mountain time (with daylight savings) then every value will be flagged as shifted since there is a systematic 1 hour shift (and two hours during daylight time).

cwhanse: I thought that the timezone of `clearsky_midday` (or `transit`) was assumed to be known. I'm not sure why a user would pass data with a different timezone. Maybe we need to talk with Matt again

wfvining: It is known, and the situation above is not likely, but I thought it was important to consider, since it is possible given that the time zone of `midday` (previously `daytime`) is not known.

wfvining:
Now that we pass midday time in minutes since midnight, the function is substantially more general, supporting the detection of shifts between arbitrary event times (`midday`) and reference times for the same events (`transit`). For example the same method could be applied to sunrise and sunset times. Should we change the parameter names and docs to reflect this use (with the caveat that it has really only been tested 'in the wild' on midday times). I imagine combining the output for transit time, sunrise time, and sunset time could produce a better result, and we may want to let users know they can easily do that with the same function.

cwhanse:
> Should we change the parameter names and docs to reflect this use (with the caveat that it has really only been tested 'in the wild' on midday times). I imagine combining the output for transit time, sunrise time, and sunset time could produce a better result, and we may want to let users know they can easily do that with the same function.

Good idea, I think.

cwhanse (approved):

PR #76, opened 2020-09-28
Geometric clipping filter from PVFleets

wfvining:
## Description

Adds a clipping detection function from PVFleets based on the slope of the data. The new function calculates a clipping threshold independently on each day.

## Checklist

### DIFF
--- a/pvanalytics/features/clipping.py
+++ b/pvanalytics/features/clipping.py
@@ -313,19 +313,19 @@ def geometric(power_ac, clip_min=0.8, daily_fraction_min=0.9,
     backward_diff = abs(power_normalized - power_normalized.shift(1))
     daily_max = _group.by_day(power_normalized).transform('max')
     fraction_of_max = power_normalized / daily_max
-    candidate_clipping = ((power_normalized >= clip_min)
-                          & ((forward_diff <= slope_max)
-                             | (backward_diff <= slope_max))
-                          & (fraction_of_max >= daily_fraction_min))
+    apparent_clipping = ((power_normalized >= clip_min)
+                         & ((forward_diff <= slope_max)
+                            | (backward_diff <= slope_max))
+                         & (fraction_of_max >= daily_fraction_min))
     # clipped values must be part of a sequence of clipped values
     # that is longer than `length_min`
-    valid_sequence = _group.run_lengths(candidate_clipping) >= length_min
-    clipping = candidate_clipping & valid_sequence
+    valid_sequence = _group.run_lengths(apparent_clipping) >= length_min
+    candidate_clipping = apparent_clipping & valid_sequence
     # Establish a clipping threshold for each day. The threshold is the
     # minimum candidate clipping value on that day (or NaN on days with
     # no clipping).
     daily_clipping_threshold = _group.by_day(
-        power_normalized[clipping].reindex_like(power_normalized)
+        power_normalized[candidate_clipping].reindex_like(power_normalized)
     ).transform('min')
     # For days with clipping, any value greater than the daily clipping
     # threshold is flagged as clipped. (Days without clipping are implicitly
--- a/pvanalytics/features/clipping.py
+++ b/pvanalytics/features/clipping.py
@@ -303,13 +303,10 @@ def geometric(power_ac, clip_min=0.8, daily_fraction_min=0.9,
     clipping = candidate_clipping & valid_sequence
     # Establish a clipping threshold for each day. The threshold is the
     # minimum value that was flagged as clipping according to the criteria
-    # above.
+    # above (or NaN on days with no clipping).
     daily_clipping_threshold = _group.by_day(
-        clipping
-    ).transform(
-        # TODO try to find a less convoluted way to do this
-        lambda day: power_normalized[day.index][day].min()
-    )
+        power_normalized[clipping].reindex_like(power_normalized)
+    ).transform('min')
     # For days with clipping (according to the criteria above), any value
     # greater than the daily clipping threshold is flagged as clipped.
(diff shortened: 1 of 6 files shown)


wfvining:
Close to being done, but the tests demand greater accuracy than the function is actually capable of. On a few days some slightly lower values that are not actually part of the simulated clipping are identified as clipped. I think this is an acceptable behavior, I just need to re-work the tests to allow for flagging values that are close to the clipping level as clipped.

pvanalytics/features/clipping.py
```diff
@@ -223,3 +226,110 @@ def threshold(ac_power, slope_max=0.0035, power_min=0.75,
...
+    # index : DatetimeIndex
+    #
+    # Returns
+    # -------
+    # float
+    #     Timestamp spacing in minutes.
+    freq = pd.infer_freq(index)
+    if freq:
+        return pd.to_timedelta(frequencies.to_offset(freq)).seconds / 60
+    return index.to_series().diff().astype('timedelta64[m]').mode()[0]
```
wfvining, line 245: I'm not sure whether we want to keep this function or just have users pass the frequency as a param when it cannot be inferred from the index. Passing an extra param would be more consistent with how we have done things in other functions ported from PVFleets

pvanalytics/features/clipping.py
```diff
@@ -223,3 +226,110 @@ def threshold(ac_power, slope_max=0.0035, power_min=0.75,
...
+        will be inferred from the index of `power_ac`.
+    derivative_max : float, optional
+        Values with a derivative less than `derivative_max` are
+        potentially clipped. If not specified the value is based on
+        the timestamp spacing in `power_ac` according to the following
+        formula:
+
+        .. math::
+
+           0.00005 * f_m + 0.0009
```
cwhanse, line 297: Maybe appears better without the '*'

pvanalytics/features/clipping.py
```diff
@@ -223,3 +226,110 @@ def threshold(ac_power, slope_max=0.0035, power_min=0.75,
...
+    power_ac.loc[power_ac < 0] = 0
+    power_normalized = _normalize.min_max(power_ac)
+    forward_diff = abs(power_normalized - power_normalized.shift(-1))
+    backward_diff = abs(power_normalized - power_normalized.shift(1))
+    daily_max = _group.by_day(power_normalized).transform('max')
+    fraction_of_max = power_normalized / daily_max
+    candidate_clipping = ((power_normalized >= clip_min)
+                          & ((forward_diff <= derivative_max)
+                             | (backward_diff <= derivative_max))
+                          & (fraction_of_max >= daily_fraction_min))
```
cwhanse, line 321: Maybe to discuss with Matt: why both `>= clip_min` and `>= daily_fraction_min`?

pvanalytics/features/clipping.py
```diff
@@ -223,3 +226,110 @@ def threshold(ac_power, slope_max=0.0035, power_min=0.75,
...
+    # clipped values must be part of a sequence of clipped values
+    # that is longer than `length_min`
+    valid_sequence = _group.run_lengths(candidate_clipping) > length_min
```
cwhanse, line 324: `>=` here, I think

pvanalytics/features/clipping.py
```diff
@@ -223,3 +226,110 @@ def threshold(ac_power, slope_max=0.0035, power_min=0.75,
...
+    clipping = candidate_clipping & valid_sequence
```
cwhanse, line 325: This is only editorial but I would call this `bool` Series `candidate_clipping`, i.e., put the `valid_sequence` filter in with the other criteria.

pvanalytics/features/clipping.py
wfvining: If we change the comparison to `>=` we also need to increase the default to 3.

pvanalytics/features/clipping.py
wfvining:
```suggestion
           0.00005 f_m + 0.0009
```

pvanalytics/features/clipping.py
wfvining: This is in flux. It is likely that the global `clip_min` parameter is not needed and will be removed.

pvanalytics/features/clipping.py
```diff
@@ -223,3 +226,110 @@ def threshold(ac_power, slope_max=0.0035, power_min=0.75,
...
+    power_ac = power_ac.copy()
+    power_ac.loc[power_ac < 0] = 0
+    power_normalized = _normalize.min_max(power_ac)
+    forward_diff = abs(power_normalized - power_normalized.shift(-1))
+    backward_diff = abs(power_normalized - power_normalized.shift(1))
+    daily_max = _group.by_day(power_normalized).transform('max')
+    fraction_of_max = power_normalized / daily_max
+    candidate_clipping = ((power_normalized >= clip_min)
+                          & ((forward_diff <= derivative_max)
+                             | (backward_diff <= derivative_max))
```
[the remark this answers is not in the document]
wfvining: This is correct. Let's add a comment to note the reason in-line.

pvanalytics/features/clipping.py
cwhanse: `>=` is more consistent with the docstring as written. Could go either way, but probably harder to describe `>`

wfvining: I think it is worth making the parameter easy to describe/understand. I was preoccupied with the value since PVFleets uses 2, but the comparison is the complement: every value less than or equal to 2 is set to `False`.

wfvining:
New version has evolved substantially. Closing this pull request in favor of #93

PR #77, opened 2020-10-05
Identify surface orientation from AC power and clearsky irradiance

wfvining:
## Description

Implements method for identifying tilt and azimuth given AC power, clear sky irradiance, and solar position.

## Checklist

### DIFF
--- a/pvanalytics/system.py
+++ b/pvanalytics/system.py
@@ -496,6 +496,8 @@ def orientation_fit_pvwatts(power_ac, ghi, dhi, dni,
     - the DC capacity of the system
     - the DC input limit of the inverter.
 
+    All parameters passed as a Series must have the same index.
+
     Parameters
     ----------
     power_ac : Series
@@ -533,6 +535,16 @@ def orientation_fit_pvwatts(power_ac, ghi, dhi, dni,
         :math:`r^2` value for the fit at the returned orientation.
 
     """
+    power_ac = power_ac.dropna()
+    ghi = ghi.dropna()
+    dhi = dhi.dropna()
+    dni = dni.dropna()
+    solar_azimuth = solar_azimuth.dropna()
+    solar_zenith = solar_zenith.dropna()
+    if isinstance(temperature, pd.Series):
+        temperature = temperature.dropna()
+    if isinstance(wind_speed, pd.Series):
+        wind_speed = wind_speed.dropna()
     if temperature_model_parameters is None:
         temperature_model_parameters = \
             TEMPERATURE_MODEL_PARAMETERS['sapm']['open_rack_glass_glass']
--- a/pvanalytics/system.py
+++ b/pvanalytics/system.py
@@ -412,41 +412,42 @@ def _power_residuals_from_clearsky(system_params,
                                    wind_speed,
                                    temperature_coefficient,
                                    temperature_model_parameters):
-    # Return the residuals between a system with parameters given in `params`
-    # and the data in `power_ac`.
-    #
-    # Parameters
-    # ----------
-    # system_params : array-like
-    #     array of four floats: tilt, azimuth, DC capacity, and inverter
-    #     DC input limit.
-    # ghi : Series
-    #     Clear sky GHI
-    # dhi : Series
-    #     Clear sky DHI
-    # dni : Series
-    #     Clear sky DNI
-    # solar_zenith : Series
-    #     Solar zenith at the same times as data in `power_ac`
-    # solar_azimuth : Series
-    #     Solar azimuth at the same times as data in `power_ac`
-    # power_ac : Series
-    #     Measured AC power under clear sky conditions.
-    # temperature : float or Series
-    #     Air temperature at which to model the hypothetical system. If a
-    #     float then a constant temperature is used. If a Series, must have
-    #     the same index as `power_ac`. [C]
-    # wind_speed : float or Series
-    #     Wind speed. If a float then a constant wind speed is used. If a
-    #     Series, must have the same index as `power_ac`. [m/s]
-    # temperature_model_parameters : dict
-    #     Parameters fot the cell temperature model.
-    #
-    # Returns
-    # -------
-    # Series
-    #     Difference between `power_ac` and the PVWatts output with the
-    #     given parameters.
+    """Return the residuals between a system with parameters given in `params`
+    and the data in `power_ac`.
+
+    Parameters
+    ----------
+    system_params : array-like
+        array of four floats: tilt, azimuth, DC capacity, and inverter
+        DC input limit.
+    ghi : Series
+        Clear sky GHI
+    dhi : Series
+        Clear sky DHI
+    dni : Series
+        Clear sky DNI
+    solar_zenith : Series
+        Solar zenith at the same times as data in `power_ac`
+    solar_azimuth : Series
+        Solar azimuth at the same times as data in `power_ac`
+    power_ac : Series
+        Measured AC power under clear sky conditions.
+    temperature : float or Series
+        Air temperature at which to model the hypothetical system. If a
+        float then a constant temperature is used. If a Series, must have
+        the same index as `power_ac`. [C]
+    wind_speed : float or Series
+        Wind speed. If a float then a constant wind speed is used. If a
+        Series, must have the same index as `power_ac`. [m/s]
+    temperature_model_parameters : dict
+        Parameters fot the cell temperature model.
+
+    Returns
+    -------
+    Series
+        Difference between `power_ac` and the PVWatts output with the
+        given parameters.
+    """
     tilt = system_params[0]
     azimuth = system_params[1]
     dc_capacity = system_params[2]
--- a/pvanalytics/system.py
+++ b/pvanalytics/system.py
@@ -412,7 +412,7 @@ def _power_residuals_from_clearsky(system_params,
                                    wind_speed,
                                    temperature_coefficient,
                                    temperature_model_parameters):
-    """Return the residuals between a system with parameters given in `params`
+    """Return the residuals between a system with parameters given in `system_params`
     and the data in `power_ac`.
 
     Parameters
@@ -439,14 +439,22 @@ def _power_residuals_from_clearsky(system_params,
     wind_speed : float or Series
         Wind speed. If a float then a constant wind speed is used. If a
         Series, must have the same index as `power_ac`. [m/s]
+    temperature_coefficient : float
+        Temperature coefficient of DC power. [1/C]
     temperature_model_parameters : dict
-        Parameters fot the cell temperature model.
+        Parameters for the cell temperature model.
 
     Returns
     -------
     Series
         Difference between `power_ac` and the PVWatts output with the
         given parameters.
+    
+    Notes
+    ------
+    Uses the defaults in `pvlib.irradiance.get_total_irradiance` to calculated plane-of-array
+    irradiance, i.e., the isotropic model for sky diffuse irradiance, and the Perez irradiance
+    transposition model.
     """
     tilt = system_params[0]
     azimuth = system_params[1]
@@ -485,9 +493,12 @@ def infer_orientation_fit_pvwatts(power_ac, ghi, dhi, dni,
                                   temperature=25, wind_speed=0,
                                   temperature_coefficient=-0.004,
                                   temperature_model_parameters=None):
-    """Get the tilt and azimuth that give pvwatts output that most closely
+    """Get the tilt and azimuth that give PVWatts output that most closely
     fits the data in `power_ac`.
 
+    Input data `power_ac`, `ghi`, `dhi`, `dni` should reflect clear-sky
+    conditions.
+    
     Uses non-linear least squares to optimize over four free variables
     to find the values that result in the best fit between power modeled
     using PVWatts and `power_ac`. The four free variables are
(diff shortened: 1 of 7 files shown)


pvanalytics/system.py
```diff
@@ -401,3 +403,172 @@ def infer_orientation_daily_peak(power_or_poa, sunny, tilts,
...
+
+
+def _power_residuals_from_clearsky(system_params,
+                                   ghi, dhi, dni,
+                                   solar_zenith, solar_azimuth,
+                                   power_ac, temperature,
+                                   wind_speed,
+                                   temperature_coefficient,
+                                   temperature_model_parameters):
+    # Return the residuals between a system with parameters given in `params`
```
wholmgren, line 415: why comment the doc string?

pvanalytics/system.py
```diff
@@ -401,3 +403,172 @@ def infer_orientation_daily_peak(power_or_poa, sunny, tilts,
...
+    #
+    # Returns
+    # -------
+    # Series
+    #     Difference between `power_ac` and the PVWatts output with the
+    #     given parameters.
+    tilt = system_params[0]
+    azimuth = system_params[1]
+    dc_capacity = system_params[2]
+    dc_limit = system_params[3]
```
wholmgren, line 453: I think `dc_inverter_limit` would be more clear. Or `pdc0_inverter`.

wholmgren (commented):
This is a great feature. 

Why restrict the function to orientation when it also performs a fit over DC capacity and inverter input limit?

wholmgren:
> Why restrict the function to orientation when it also performs a fit over DC capacity and inverter input limit?

Thinking about this some more, the capacity values are relatively sensitive to temperature model/coefficient, AOI model, and spectral model, while orientation is largely independent of those choices. Expanding the API to include all of those options could be a real mess, so perhaps a good idea to restrict the scope.

pvanalytics/system.py
```diff
@@ -401,3 +403,172 @@ def infer_orientation_daily_peak(power_or_poa, sunny, tilts,
...
+    power_ac = power_ac.dropna()
+    ghi = ghi.dropna()
+    dhi = dhi.dropna()
+    dni = dni.dropna()
+    solar_azimuth = solar_azimuth.dropna()
+    solar_zenith = solar_zenith.dropna()
+    if isinstance(temperature, pd.Series):
+        temperature = temperature.dropna()
+    if isinstance(wind_speed, pd.Series):
+        wind_speed = wind_speed.dropna()
```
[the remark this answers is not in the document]
wfvining: Maybe it would be better to require users to remove NAs and align the indices before calling this?

wfvining:
I originally restricted to only the tilt and azimuth because I didn't know if the other parameters would be accurate (sounds like they might not be). There is, however, evidence in the literature that this type of approach can work for tilt and azimuth (Haghdadi N; Copper J; Bruce A; MacGill I, 2017, 'A method to estimate the location and orientation of distributed photovoltaic systems from their generation output data', Renewable Energy, vol. 108, pp. 390 - 400, http://dx.doi.org/10.1016/j.renene.2017.02.080). I would like to eventually push towards the method described in that paper, and this function seems like a good first step in that direction.

pvanalytics/system.py
```diff
@@ -401,3 +403,172 @@ def infer_orientation_daily_peak(power_or_poa, sunny, tilts,
...
+    # Calculate the coefficient of determination from `residuals`
+    model = data + residuals
+    _, _, r, _, _ = scipy.stats.linregress(model, data)
+    return r*r
+
+
+def orientation_fit_pvwatts(power_ac, ghi, dhi, dni,
+                            solar_zenith, solar_azimuth,
+                            temperature=25, wind_speed=0,
+                            temperature_coefficient=-0.002,
```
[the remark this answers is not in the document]
wfvining: I think I just pulled that from the documentation for `pvlib.pvsystem.pvwatts_dc()`. Do you have a recommendation for a better default, I don't really have enough familiarity to say what a good default would be here?

pvanalytics/system.py
wfvining: I've generally followed the convention that only functions in the public API of a module have docstrings, private functions get block comments. I think PEP-8 hints at this, but doesn't demand one way or another. 

I think I mainly do it because I sometimes use another language where you get a compiler warning if you attach documentation to a non-public function. Do you have a preference?

wholmgren: I prefer regular docstrings so that my IDE renders it correctly. 

PEP 8 says "Docstrings are not necessary for non-public methods, but you should have a comment that describes what the method does." I interpret that as "if you're not going to write a doc string for your private function, at least make a brief comment."

pvanalytics/system.py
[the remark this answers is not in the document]
wholmgren: that docstring says "Typically -0.002 to -0.005 per degree C." Maybe use the average?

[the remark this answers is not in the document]
cwhanse: Since PVWatts doesn't use a linear temperature model, I'd suggest using -0.004 as the default. Would be interesting to calculate the slope of a linear approximation to the Fuentes model with PVWatts defaults. Someday.

pvanalytics/system.py
wfvining: That's a reasonable interpretation. I actually prefer comments to docstrings for the same reason. My docstring highlighting color is brighter and I like not having private functions jump out at me when I'm reading through unfamiliar code. I'm fine with changing it, since I suspect I'm in the minority here.

pvanalytics/system.py
```diff
@@ -401,3 +403,172 @@ def infer_orientation_daily_peak(power_or_poa, sunny, tilts,
...
+        wind_speed,
+        **temperature_model_parameters
+    )
+    pdc = pvlib.pvsystem.pvwatts_dc(
+        poa['poa_global'],
+        temp_cell,
+        dc_capacity,
+        temperature_coefficient
+    )
+    return power_ac - pvlib.inverter.pvwatts(pdc, dc_limit)
```
wholmgren, line 472: `pvlib.inverter.pvwatts` will bump the minimum pvlib to 0.8. I ran into this at runtime when trying to test out the code.

wfvining: Thanks. I will update the requirements.

pvanalytics/system.py
```diff
@@ -401,3 +403,177 @@ def infer_orientation_daily_peak(power_or_poa, sunny, tilts,
...
+
+
+def _power_residuals_from_clearsky(system_params,
+                                   ghi, dhi, dni,
+                                   solar_zenith, solar_azimuth,
+                                   power_ac, temperature,
+                                   wind_speed,
+                                   temperature_coefficient,
+                                   temperature_model_parameters):
+    """Return the residuals between a system with parameters given in `params`
```
cwhanse, line 415:
```suggestion
    """Return the residuals between a system with parameters given in `system_params`
```

pvanalytics/system.py
```diff
@@ -401,3 +403,177 @@ def infer_orientation_daily_peak(power_or_poa, sunny, tilts,
...
+        Clear sky GHI
+    dhi : Series
+        Clear sky DHI
+    dni : Series
+        Clear sky DNI
+    solar_zenith : Series
+        Solar zenith at the same times as data in `power_ac`
+    solar_azimuth : Series
+        Solar azimuth at the same times as data in `power_ac`
+    power_ac : Series
```
cwhanse, line 433: Purely editorial, better if `power_ac` precedes `solar_zenith` and `solar_azimuth`

pvanalytics/system.py
```diff
@@ -401,3 +403,177 @@ def infer_orientation_daily_peak(power_or_poa, sunny, tilts,
...
+        Measured AC power under clear sky conditions.
+    temperature : float or Series
+        Air temperature at which to model the hypothetical system. If a
+        float then a constant temperature is used. If a Series, must have
+        the same index as `power_ac`. [C]
+    wind_speed : float or Series
+        Wind speed. If a float then a constant wind speed is used. If a
+        Series, must have the same index as `power_ac`. [m/s]
+    temperature_model_parameters : dict
+        Parameters fot the cell temperature model.
```
cwhanse, line 443:
```suggestion
        Parameters for the cell temperature model.
```

pvanalytics/system.py
```diff
@@ -401,3 +403,177 @@ def infer_orientation_daily_peak(power_or_poa, sunny, tilts,
...
+    power_ac : Series
+        Measured AC power under clear sky conditions.
+    temperature : float or Series
+        Air temperature at which to model the hypothetical system. If a
+        float then a constant temperature is used. If a Series, must have
+        the same index as `power_ac`. [C]
+    wind_speed : float or Series
+        Wind speed. If a float then a constant wind speed is used. If a
+        Series, must have the same index as `power_ac`. [m/s]
+    temperature_model_parameters : dict
```
cwhanse, line 442:
```suggestion
    temperature_coefficient : float
        Temperature coefficient of DC power. [1/C]
    temperature_model_parameters : dict
```

pvanalytics/system.py
```diff
@@ -401,3 +403,177 @@ def infer_orientation_daily_peak(power_or_poa, sunny, tilts,
...
+        Parameters fot the cell temperature model.
+
+    Returns
+    -------
+    Series
+        Difference between `power_ac` and the PVWatts output with the
+        given parameters.
+    """
```
cwhanse, line 450:
```suggestion
    
    Notes
    ------
    Uses the defaults in `pvlib.irradiance.get_total_irradiance` to calculated plane-of-array
    irradiance, i.e., the isotropic model for sky diffuse irradiance, and the Perez irradiance
    transposition model.
    """
```

pvanalytics/system.py
```diff
@@ -401,3 +403,177 @@ def infer_orientation_daily_peak(power_or_poa, sunny, tilts,
...
+    return r*r
+
+
+def infer_orientation_fit_pvwatts(power_ac, ghi, dhi, dni,
+                                  solar_zenith, solar_azimuth,
+                                  temperature=25, wind_speed=0,
+                                  temperature_coefficient=-0.004,
+                                  temperature_model_parameters=None):
+    """Get the tilt and azimuth that give pvwatts output that most closely
+    fits the data in `power_ac`.
```
cwhanse, line 489:
```suggestion
    """Get the tilt and azimuth that give PVWatts output that most closely
    fits the data in `power_ac`.
```

pvanalytics/system.py
```diff
@@ -401,3 +403,177 @@ def infer_orientation_daily_peak(power_or_poa, sunny, tilts,
...
+
+    Uses non-linear least squares to optimize over four free variables
```
cwhanse, line 491:
```suggestion
    Input data `power_ac`, `ghi`, `dhi`, `dni` should reflect clear-sky
    conditions.
    
    Uses non-linear least squares to optimize over four free variables
```

pvanalytics/system.py
```diff
@@ -401,3 +403,177 @@ def infer_orientation_daily_peak(power_or_poa, sunny, tilts,
...
+    All parameters passed as a Series must have the same index and must not
+    contain any undefined values (i.e. NaNs). If any input contains NaNs a
+    ValueError is raised.
+
+    Parameters
+    ----------
+    power_ac : Series
+        AC power from the system in clear sky conditions.
+    ghi : Series
+        Clear sky GHI with same index as `power_ac`.
```
cwhanse, line 509:
```suggestion
        Clear sky GHI with same index as `power_ac`. [W/m^2]
```

pvanalytics/system.py
```diff
@@ -401,3 +403,177 @@ def infer_orientation_daily_peak(power_or_poa, sunny, tilts,
...
+    dhi : Series
+        Clear sky DHI with same index as `power_ac`.
```
cwhanse, line 511:
```suggestion
        Clear sky DHI with same index as `power_ac`. [W/m^2]
```

pvanalytics/system.py
```diff
@@ -401,3 +403,177 @@ def infer_orientation_daily_peak(power_or_poa, sunny, tilts,
...
+    dni : Series
+        Clear sky DNI with same index as `power_ac`.
```
cwhanse, line 513:
```suggestion
        Clear sky DNI with same index as `power_ac`. [W/m^2]
```

pvanalytics/system.py
```diff
@@ -401,3 +403,177 @@ def infer_orientation_daily_peak(power_or_poa, sunny, tilts,
...
+    """Get the tilt and azimuth that give pvwatts output that most closely
+    fits the data in `power_ac`.
+
+    Uses non-linear least squares to optimize over four free variables
+    to find the values that result in the best fit between power modeled
+    using PVWatts and `power_ac`. The four free variables are
+
+    - surface tilt
+    - surface azimuth
+    - the DC capacity of the system
```
cwhanse, line 497: May be worth a remark why DC capacity and DC input limit are not returned - I'm guessing that the algorithm doesn't reliably estimate the values?

cwhanse: Also, if the algorithm can handle `power_ac` from systems with inverter clipping (it appears that it could) that's worth a remark.

pvanalytics/system.py
wfvining: I think I'd like this to be either first or last, would you be okay with `power_ac`, `ghi`, ..., `solar_zenith`?

cwhanse: Yes

wfvining: On second thought after the clear sky irradiance is fine. Not really important where it goes, since it is only every used as a kwarg by `scipy.optimize.least_squares`.

cwhanse (approved):

pvanalytics/system.py
kanderso-nrel: @cwhanse I don't think this code is actually using Perez.  Is that something we should update?  Seems like a shame to deviate from PVWatts in a way that will presumably make the fits worse.

cc @kperrynrel since she's getting bad fits :)

kanderso-nrel: Alternatively, should we switch the code to use Perez?  I think the extra inputs could be calculated automatically.  It would be more aligned with PVWatts itself, plus hopefully return better results.

kanderso-nrel: Fair point.  Note that `ModelChain.with_pvwatts` currently uses the Sandia temperature model too: https://github.com/pvlib/pvlib-python/blob/master/pvlib/modelchain.py#L49-L59

I'll open an issue.

kanderso-nrel: #147

PR #79, opened 2020-10-12
Look for daylight savings shifts across specific dates

### LINKED ISSUE #66: determine if a time series is DST aware or not
I'd like a function that tells me if a time series uses a fixed offset timezone like `'Etc/GMT+7'` or a DST aware timezone like `'America/Denver'`.

In the PV world, we often don't have data that would occur right at the DST transition (2 am local). So my suggestion is to find the sunrise, sunset, and/or solar noon in the ~7 days before and after a DST transition. Then determine if there's a 1 hour shift in that data.

wfvining:
## Description

Given the dates on which daylight savings transitions occur, determine whether a series has DST by testing for a large shift between sunrise time in the week before and after the shift date.

## Checklist

- [x] Closes #66

### DIFF
--- a/pvanalytics/quality/time.py
+++ b/pvanalytics/quality/time.py
@@ -201,7 +201,7 @@ def dst_dates(dates, tz):
     Returns
     -------
     Series
-        Boolean Series with True on days where daylight savins transitions
+        Boolean Series with True on days where daylight savings transitions
         occur.
     """
     midnight = pd.Series(
@@ -263,8 +263,9 @@ def has_dst(events, tz, window=7, min_difference=45):
     events : Series
         Series with one timestamp for each day. The timestamp should
         correspond to an event that occurs at roughly the same time on
-        each day, and shifts with daylight savings transitions. For example,
-        you may pass sunrise, sunset, or solar transit time.
+        each day. For example,
+        you may pass sunrise, sunset, or solar transit time. `events` need
+        not be localized.
     tz : str
         Name of a timezone that observes daylight savings and has the same
         or similar UTC offset as the expected time zone for `events`.
--- a/pvanalytics/quality/time.py
+++ b/pvanalytics/quality/time.py
@@ -228,7 +228,7 @@ def _has_dst(events, date, window, min_difference):
     window : Timedelta
         Number of days before and after `date` to use when checking for
         for shifts in `events`
-    min_different : int
+    min_difference : int
         Minimum difference between mean event time before `date` and mean
         event time after `date` for a shift tom be detected.
 
--- a/pvanalytics/quality/time.py
+++ b/pvanalytics/quality/time.py
@@ -186,9 +186,9 @@ def shifts_ruptures(event_times, reference_times, period_min=2,
     return shift_amount != 0, shift_amount
 
 
-def _has_dst(event_times, date, window, min_difference):
-    before = event_times[date - window:date - pd.Timedelta(days=1)]
-    after = event_times[date:date + window]
+def _has_dst(events, date, window, min_difference):
+    before = events[date - window:date - pd.Timedelta(days=1)]
+    after = events[date:date + window]
     if len(before) == 0 or len(after) == 0:
         raise ValueError(f"Insufficient data at {date}.")
     before = before.dt.hour * 60 + before.dt.minute
@@ -196,8 +196,8 @@ def _has_dst(event_times, date, window, min_difference):
     return abs(before.mean() - after.mean()) > min_difference
 
 
-def has_dst(event_times, tz, shift_dates=None, window=7, min_difference=45):
-    """Return true if `daynight_mask` appears to have daylight-savings shifts
+def has_dst(events, tz, window=7, min_difference=45):
+    """Return true if `events` appears to have daylight-savings shifts
     at the dates in `shift_dates`.
 
     Compares the mean event time in minutes since midnight over the
@@ -207,17 +207,17 @@ def has_dst(event_times, tz, shift_dates=None, window=7, min_difference=45):
 
     Parameters
     ----------
-    event_times : Series
+    events : Series
         Series with one timestamp for each day. The timestamp should
         correspond to an event that occurs at roughly the same time on
         each day, and shifts with daylight savings transitions. For example,
         you may pass sunrise, sunset, or solar transit time.
     tz : str
         Name of a timezone that observes daylight savings and has the same
-        or similar UTC offset as the expected time zone for `event_times`.
+        or similar UTC offset as the expected time zone for `events`.
     window : int, default 7
         Number of days before and after the shift date to consider. When
-        passing rounded timestamps in `event_times` it may be necessary to
+        passing rounded timestamps in `events` it may be necessary to
         use a smaller window. [days]
     min_difference : int, default 45
         Minimum difference between the mean event time before the shift
@@ -227,21 +227,23 @@ def has_dst(event_times, tz, shift_dates=None, window=7, min_difference=45):
 
     Returns
     -------
-    list of bool
-        Boolean indicating whether a DST shift was found at each date in
-        `shift_dates`.
+    Series
+        Boolean Series indexed by dates on which Daylight Savings transitions
+        are expected to occur in `tz` and True if a shift was found on that
+        date in `events`
 
     Raises
     ------
     ValueError
         If there is no data before or after a shift date or there are no
         daylight-savings shifts in `tz` for the dates covered by
-        `event_times`.
+        `events`.
     """
     # Build a timestamp at noon on each day in the data
-    s = pd.Series(pd.DatetimeIndex(
-        event_times.index.tz_localize(None).date),
-        index=event_times.index)
+    s = pd.Series(
+        pd.DatetimeIndex(events.index.tz_localize(None).date),
+        index=events.index
+    )
     s = s + pd.Timedelta(hours=12)
     s = s.dt.tz_localize(tz)
     dst_shift = s.apply(lambda t: t.tzinfo.dst(t).total_seconds() / 3600)
@@ -250,11 +252,11 @@ def has_dst(event_times, tz, shift_dates=None, window=7, min_difference=45):
         raise ValueError("No daylight savings shifts in expected "
                          f"timezone ({tz}) on dates in input")
     window = pd.Timedelta(days=window)
-    event_times = event_times.dropna()
+    events = events.dropna()
     return shift_dates.apply(
         lambda t: _has_dst(
-            event_times,
-            pd.Timestamp(t.date(), tz=event_times.index.tz),
+            events,
+            pd.Timestamp(t.date(), tz=events.index.tz),
             window,
             min_difference
         )
--- a/pvanalytics/quality/time.py
+++ b/pvanalytics/quality/time.py
@@ -196,7 +196,7 @@ def _has_dst(event_times, date, window, min_difference):
     return abs(before.mean() - after.mean()) > min_difference
 
 
-def has_dst(event_times, shift_dates, window=7, min_difference=45):
+def has_dst(event_times, tz, shift_dates=None, window=7, min_difference=45):
     """Return true if `daynight_mask` appears to have daylight-savings shifts
     at the dates in `shift_dates`.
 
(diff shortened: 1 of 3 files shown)


wfvining:
Needs a some cleaning up RE documentation variable names as well as more testing (i.e. what should the function do with missing data).

pvanalytics/quality/time.py
```diff
@@ -184,3 +184,63 @@ def shifts_ruptures(event_times, reference_times, period_min=2,
...
+    min_difference : int, default 45
+        Minimum difference between the mean event time before the shift
+        date and the mean event time after the event time. If the difference
+        is greater than `min_difference` a shift has occurred on that date.
+        [minutes]
+
+    Returns
+    -------
+    list of bool
+        Boolean indicating whether a DST shift was found at each date in
```
cwhanse, line 232: should this Bool series be indexed by `event_times` rather than `shift_dates`? That's more a question about user expectation

pvanalytics/quality/time.py
```diff
@@ -184,3 +184,63 @@ def shifts_ruptures(event_times, reference_times, period_min=2,
...
+    at the dates in `shift_dates`.
+
+    Compares the mean event time in minutes since midnight over the
+    `window` days before and after each date in `shift_dates`. If the
+    difference is greater than `min_difference` then a shift has occurred
+    on that date.
+
+    Parameters
+    ----------
+    event_times : Series
```
cwhanse, line 210: not wild about `event_times` or `shift_dates`. Maybe `events`? Maybe `times_to_test`?  For `shift_dates` maybe `dst_shift_dates`?

cwhanse (commented):
@wholmgren interested in your view of `has_dst`, if it meets the need you expressed

wholmgren (commented):
I want a function that takes a time series of AC power and tells me if it's DST aware or not. Perhaps there's a way to combine this function with other pvanalytics functions to accomplish that, but I've not closely watched the library development. The `event_times` argument leaves a lot of room for error because if I can't trust my time stamps then I can't trust my solar position.

Is there some way to use pytz to look up the `shift_dates`? I'm not too thrilled about the idea of looking up that date every year. Is there really a need for `window` and `min_difference`? 

Try testing against pvdaq data since that's where I ran into this problem. Some ids to start with include 4, 10, 1199, 1289

wfvining:
@wholmgren You can use other PVAnalytics functions to get a day/night mask. From there it is simple to get approximate sunrise/sunset/transit times.

> Is there some way to use pytz to look up the shift_dates? I'm not too thrilled about the idea of looking up that date every year.

I made this a user input since `pytz` doesn't provide an actual API function or field for getting these dates. They are accessible via private fields (can't remember the name right now), but I didn't want to use that in pvanalytics directly. Maybe it isn't actually that big of a deal.

cwhanse:
Since 2007, daylight savings has been from second week of March to first week of November. Prior to 2007, it was first week of April to last week of October. Maybe in combination with `window` we can build the daylight savings transition periods into the function.

wfvining:
> Since 2007, daylight savings has been from second week of March to first week of November. Prior to 2007, it was first week of April to last week of October. Maybe in combination with window we can build the daylight savings transition periods into the function.

Is this true globally? I thought this was only in the US.

cwhanse:
> Is this true globally? I thought this was only in the US.

Good point, that's for the US. 2020 times [here](https://www.timeanddate.com/time/dst/2020.html).

wfvining:
My thinking now is to have users pass a timezone and use that to generate a timestamp for each day at noon on the same dates as the sunrise timestamps. From that we can use `.tzinfo().dst(timestamp)` to get the "correct" DST shift for each day. Extracting the dates from that is just a matter of searching for the days on which the DST timedelta is different from the preceding day. Is this similar to what you were thinking @wholmgren?

wholmgren:
So the user would pass a candidate DST aware time zone and function would say if it's correct or not? I think that could work.

cwhanse:
@wfvining I can review but want to know if the PR is up to date, with the conversation here. I don't see a time zone input, so I'm not sure.

wfvining:
@cwhanse I just pushed it. It was somewhat trickier than I expected.

pvanalytics/quality/time.py
wfvining: Now returns series indexed by daylight savings transition dates. I'm not 100% sold in that being the right behavior, but I think it is the best option. To determine if the series has dst at all, users can use `time.has_dst(x).any()`.

pvanalytics/quality/time.py
```diff
@@ -184,3 +184,103 @@ def shifts_ruptures(event_times, reference_times, period_min=2,
...
+    if len(before) == 0 or len(after) == 0:
+        raise ValueError(f"Insufficient data at {date}.")
+    before = before.dt.hour * 60 + before.dt.minute
+    after = after.dt.hour * 60 + after.dt.minute
+    return abs(before.mean() - after.mean()) > min_difference
+
+
+def has_dst(events, tz, window=7, min_difference=45):
+    """Return true if `events` appears to have daylight-savings shifts
+    at the dates in `shift_dates`.
```
wfvining, line 224:
```suggestion
    """Return true if `events` appears to have daylight-savings shifts
    on the dates expected in `tz`.
```

pvanalytics/quality/time.py
```diff
@@ -184,3 +184,103 @@ def shifts_ruptures(event_times, reference_times, period_min=2,
...
+
+    Compares the mean event time in minutes since midnight over the
+    `window` days before and after each date in `shift_dates`. If the
+    difference is greater than `min_difference` then a shift has occurred
+    on that date.
```
wfvining, line 229: Rewrite.

pvanalytics/quality/time.py
```diff
@@ -184,3 +184,103 @@ def shifts_ruptures(event_times, reference_times, period_min=2,
...
+    """
+    # Build a timestamp at noon on each day in the data
+    s = pd.Series(
+        pd.DatetimeIndex(events.index.tz_localize(None).date),
+        index=events.index
+    )
+    s = s + pd.Timedelta(hours=12)
+    s = s.dt.tz_localize(tz)
+    dst_shift = s.apply(lambda t: t.tzinfo.dst(t).total_seconds() / 3600)
+    shift_dates = s[dst_shift.diff().fillna(0) != 0]
```
wfvining, line 273: Factor out to its own function: given an index and a timezone return a series (or list?) with dates of the daylight savings transitions.

pvanalytics/quality/time.py
```diff
@@ -184,3 +184,103 @@ def shifts_ruptures(event_times, reference_times, period_min=2,
...
+        date and the mean event time after the event time. If the difference
+        is greater than `min_difference` a shift has occurred on that date.
+        [minutes]
+
+    Returns
+    -------
+    Series
+        Boolean Series indexed by dates on which Daylight Savings transitions
+        are expected to occur in `tz` and True if a shift was found on that
+        date in `events`
```
wfvining, line 256: Return series with same index as `events`. All False except the transition dates which are True if a dst-like transition was observed across that date.

wfvining:
Reimplemented to return Series with same index as input, False on all days except DST transition dates which may or may not be True. Still need to do a little more thinking on how to handle missing data. Raising a `ValueError` might be too aggressive and could make the function less useful, but I don't want to quietly return made up data. Ruminating on whether to just `warn` or provide an option that will allow users to specify warning vs exception (some pandas functions provide an argument such as this).

pvanalytics/quality/time.py
```diff
@@ -184,3 +184,123 @@ def shifts_ruptures(event_times, reference_times, period_min=2,
...
+    ----------
+    dates : DatetimeIndex
+        A datetime index with one timestamp per day.
+    tz : str
+        A timezone string (e.g. 'America/Denver' or 'CET')
+
+    Returns
+    -------
+    Series
+        Boolean Series with True on days where daylight savins transitions
```
cwhanse, line 204:
```suggestion
        Boolean Series with True on days where daylight savings transitions
```

pvanalytics/quality/time.py
```diff
@@ -184,3 +184,123 @@ def shifts_ruptures(event_times, reference_times, period_min=2,
...
+
+def has_dst(events, tz, window=7, min_difference=45):
+    """Return True if `events` appears to have daylight savings shifts
+    at the dates on which `tz` transitions to or from daylight savings
+    time.
+
+    Compares the mean event time in minutes since midnight over the
+    `window` days before and after each date in `shift_dates`. If the
+    difference is greater than `min_difference` then a shift has occurred
+    on that date.
```
cwhanse, line 259:
```suggestion
    The mean event time in minutes since midnight is calculated
    over the `window` days before and after each date in `shift_dates`.
    For each date, the two mean event times (before and after) are
    compared, and if the difference is greater than `min_difference`
    then a shift has occurred on that date.
```

pvanalytics/quality/time.py
```diff
@@ -184,3 +184,123 @@ def shifts_ruptures(event_times, reference_times, period_min=2,
...
+
+    Parameters
+    ----------
+    events : Series
+        Series with one timestamp for each day. The timestamp should
+        correspond to an event that occurs at roughly the same time on
+        each day, and shifts with daylight savings transitions. For example,
```
cwhanse, line 266:
```suggestion
        each day. For example,
```
I don't think it's required that the event timestamp changes with daylight savings, that would imply that `events` is localized to a daylight savings aware timezone.

Maybe add `events need not be localized`

pvanalytics/quality/time.py
```diff
@@ -184,3 +184,123 @@ def shifts_ruptures(event_times, reference_times, period_min=2,
...
+    on `date`.
+
+    Parameters
+    ----------
+    events : Series
+        Series of timestamps, one per day.
+    date : Timestamp
+        Timestamp corresponding to midnight on the day to check for daylight
+        savings shifts.
+    window : int
```
cwhanse, line 228: As used, `window` is a Timedelta I think.

cwhanse (approved):

pvanalytics/quality/time.py
wfvining: That's true. I think I was trying to communicate that the event whose time is being recorded should occur at roughly the same UTC time each day, and therefore be subject to daylight savings shifts when recorded in local time (i.e. you could pass a series with noon as the event, but it wouldn't do any good because that event shifts with DST).

`events need not be localized` plus the examples of what kind of events we expect is probably sufficient.

pvanalytics/quality/time.py
wfvining: Yes, it is an int in the API funciton, but a Timedelta here. Thanks, good catch.

pvanalytics/quality/time.py
```diff
@@ -184,3 +184,123 @@ def shifts_ruptures(event_times, reference_times, period_min=2,
...
+    difference is greater than `min_difference` then a shift has occurred
+    on that date.
+
+    Parameters
+    ----------
+    events : Series
+        Series with one timestamp for each day. The timestamp should
+        correspond to an event that occurs at roughly the same time on
+        each day, and shifts with daylight savings transitions. For example,
+        you may pass sunrise, sunset, or solar transit time.
```
wfvining, line 267:
```suggestion
        you may pass sunrise, sunset, or solar transit time. `events` need
        not be localized.
```

pvanalytics/quality/time.py
```diff
@@ -184,3 +184,124 @@ def shifts_ruptures(event_times, reference_times, period_min=2,
...
+    on `date`.
+
+    Parameters
+    ----------
+    events : Series
+        Series of timestamps, one per day.
+    date : Timestamp
+        Timestamp corresponding to midnight on the day to check for daylight
+        savings shifts.
+    window : int
```
wfvining, line 228:
```suggestion
    window : Timedelta
```

wfvining:
@cwhanse What do you think we should do if the DST transition date is not in the index of `events`? The current behavior is to look at the first date following the transition and see if there is a shift there, but I am not sure that is the correct approach.

cwhanse:
> @cwhanse What do you think we should do if the DST transition date is not in the index of `events`? The current behavior is to look at the first date following the transition and see if there is a shift there, but I am not sure that is the correct approach.

I think either of these behaviors is consistent with the function's intent:
- return True only for dates in `events` that have DST shifts.
- fill in `events` with missing dates between the first and last date in `events.

Mild preference for the first, so that the returned Series aligns with the user's data.

PR #81, opened 2020-10-23
Use scipy.stats.median_abs_deviation in outliers.hampel

### LINKED ISSUE #80: scipy.stats.median_absolute_deviation is deprecated
Getting a warning from the tests for `pvanalytics.quality.outliers`

### From the deprecation warning:

use `median_abs_deviation` instead!
  
  To preserve the existing default behavior, use `scipy.stats.median_abs_deviation(..., scale=1/1.4826)`. The value 1.4826 is not numerically precise for scaling with a normal distribution. For a numerically precise value, use `scipy.stats.median_abs_deviation(..., scale='normal')`.

wfvining:
scipy.stats.median_absolute_deviation is deprecated in version 1.5

## Checklist

- [x] Closes #80 

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
 numpy>=1.16.0
 pandas>=0.25.0,<1.1.0
 pvlib>=0.8.0
-scipy>=1.3.0
+scipy>=1.5.0
 ruptures[optional]
--- a/pvanalytics/quality/outliers.py
+++ b/pvanalytics/quality/outliers.py
@@ -95,6 +95,7 @@ def hampel(data, window=5, max_deviation=3.0, scale=1.4826):
     median = data.rolling(window=window, center=True).median()
     deviation = abs(data - median)
     mad = data.rolling(window=window, center=True).apply(
-        stats.median_absolute_deviation
+        stats.median_abs_deviation,
+        kwargs={'scale': 'normal'}
     )
-    return deviation > max_deviation * scale * mad
+    return deviation > max_deviation * mad * scale
--- a/setup.py
+++ b/setup.py
@@ -34,7 +34,7 @@ INSTALL_REQUIRES = [
     'numpy >= 1.16.0',
     'pandas >= 0.25.1',
     'pvlib >= 0.8.0',
-    'scipy >= 1.3.0'
+    'scipy >= 1.5.0'
 ]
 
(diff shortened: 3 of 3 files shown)


cwhanse (approved):

requirements.txt
```diff
@@ -1,5 +1,5 @@
 numpy>=1.16.0
 pandas>=0.25.0,<1.1.0
 pvlib>=0.8.0
-scipy>=1.3.0
+scipy>=1.5.0
```
cwhanse, line 4: Maybe think about this for a moment - pvlib is on scipy >= 1.2.0, and it's a guiding principle here that pvanalytics and pvlib play nice together. Do we want to bump the scipy version? Or, program around the deprecation warning?

cwhanse (commented):
Let's think about the scipy bump first

requirements.txt
wfvining: PVLib tests pass at scipy 1.5.2. I'm personally okay with the different requirement, so long as it satisfies the >= cirteria. We already depend on scipy > 1.2 for the `median_absolute_deviation` function.

wholmgren: As someone that recently started trying to use pvanalytics alongside other code, I have two comments:

1. this is a big jump to a newish scipy version (4 months), and would make me think twice about installing pvanalytics in some places and prevent it in others. 
2. it would help to know the pvanalytics policy for dependencies. My suggestion is following [NEP-29](https://numpy.org/neps/nep-0029-deprecation_policy.html) for the core pydata packages. I'm not sure what to recommend for a pvlib version policy.

wfvining: Thanks @wholmgren. I was not aware of NEP-29, and I certainly see the benefits for a predictable approach to dependency management. My approach has been ad-hoc thus far, so I would not be opposed to adopting the policy laid out in NEP-29: all minor versions back 2 years.

That will mean a bit of re-working since some versions have crept higher than that would allow. For instance, to support scipy==1.2 we will need to add `statsmodels` as a dependency and use `statsmodels.robust.scale.mad()` here.

cwhanse: Ok to add statsmodels. As far as the median absolute deviation statistic, an imprecise value is good enough here.

wfvining:
Closing in favor of #85.

Issue #83, opened 2020-10-26
Adopt dependency management policy


wfvining:
[NEP-29](https://numpy.org/neps/nep-0029-deprecation_policy.html) is probably the one to use (see #81). We will support minor python versions going back 42 months (currently python >= 3.6) and numpy/scipy released in the previous 24 months (currently scipy >= 1.2 and numpy >= 1.15). 

It might also a good idea to apply this to the pandas version as well, meaning we would support pandas >= 0.23 (and < 1.1 still, see #82).

PR #85, opened 2020-10-26
Use statsmodels for maximum absolute deviation

### LINKED ISSUE #80: scipy.stats.median_absolute_deviation is deprecated

wfvining:
Rather than `scipy.stats.maximum_absolute_deviation()` we will use `statsmodels.robust.scale.mad()` to calculate the Maximum Absolute Deviation. This change will let us lower the minimum scipy version to 1.2 `scipy.stats.maximum_absolute_deviation()` was introduces in 1.3 (see #83).

## Checklist

- [x] Closes #80 

### DIFF
--- a/pvanalytics/quality/outliers.py
+++ b/pvanalytics/quality/outliers.py
@@ -60,7 +60,7 @@ def zscore(data, zmax=1.5):
     return pd.Series((abs(stats.zscore(data)) > zmax), index=data.index)
 
 
-def hampel(data, window=5, max_deviation=3.0, scale=1.4826):
+def hampel(data, window=5, max_deviation=3.0, scale=None):
     r"""Identify outliers by the Hampel identifier.
 
     The Hampel identifier is computed according to [1]_.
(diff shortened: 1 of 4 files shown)


pvanalytics/quality/outliers.py
```diff
@@ -95,6 +96,7 @@ def hampel(data, window=5, max_deviation=3.0, scale=1.4826):
     median = data.rolling(window=window, center=True).median()
     deviation = abs(data - median)
     mad = data.rolling(window=window, center=True).apply(
-        stats.median_absolute_deviation
+        robust.scale.mad,
+        kwargs={'c': 1.0/scale}
```
cwhanse, line 100: Commenting here: let's go with scale=None as the default, and not pass 'c' if scale=None.

cwhanse: For the docstring, suggest
```
Scale factor used to estimate the standard deviation as
:math:`MAD / scale`. If scale=None (default), then the scale
factor is taken to be `scipy.stats.norm.ppf(3/4.)` (approx. 0.6745),
and :math:`MAD / scale` approximates the standard deviation
of Gaussian distributed data.

cwhanse (approved):

pvanalytics/quality/outliers.py
wfvining: Sounds good.

PR #89, opened 2020-10-30
Apply NEP-29 guidelines for dependencies

### LINKED ISSUE #83: Adopt dependency management policy

wfvining:
Lowers minimum required version of pandas and numpy. Adds a CI job to test against the minimum version for all dependencies and fixes one bug related to the lower pandas version.

## Checklist

- [x] Closes #83 

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -76,6 +76,7 @@ It supports:
   and at minimum the two latest minor versions.
 - All minor versions of numpy released in the 24 months prior to the project,
   and at minimum the last three minor versions
+- The latest release of `PVLib <https://pvlib-python.readthedocs.io>`_.
 
 PVAnalytics depends on the following packages:
 
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -68,7 +68,9 @@ library status.
 Dependencies
 ------------
 
-This project follows the guidelines laid out in NEP-29. It supports:
+This project follows the guidelines laid out in
+`NEP-29 <https://numpy.org/neps/nep-0029-deprecation_policy.html>`_.
+It supports:
 
 - All minor versions of Python released 42 months prior to the project,
   and at minimum the two latest minor versions.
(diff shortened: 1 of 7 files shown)


wfvining:
@wholmgren does this satisfy your needs? I appreciate your suggestion on this. Going through the process of standardizing the dependency versions has done a lot to improve the library.

docs/index.rst
```diff
@@ -65,6 +65,20 @@ library status.
 - :py:mod:`dataclasses` contains classes for normalizing data (e.g. an
   ``IVCurve`` class)
 
+Dependencies
+------------
+
+This project follows the guidelines laid out in NEP-29. It supports:
```
wholmgren, line 71: link to NEP-29?

docs/index.rst
```diff
@@ -65,6 +65,20 @@ library status.
...
+
+- All minor versions of Python released 42 months prior to the project,
+  and at minimum the two latest minor versions.
+- All minor versions of numpy released in the 24 months prior to the project,
+  and at minimum the last three minor versions
+
```
wholmgren, line 77: might be worth specifying a pvlib dependency policy, even if it's not set in stone.

wholmgren (commented):
@wfvining yes thank you

docs/index.rst
wfvining: Not sure what the best policy is for pvlib. So far I have tended to track the latest version pretty closely rather than try to support across minor versions with breaking API differences.

cwhanse: For now I think dependency on the latest release of pvlib (v0.8) is okay. We can test pvanalytics for pvlib >= v0.8 and forward.

PR #93, opened 2020-11-06
Geometric clipping

wfvining:
New clipping filter designed by Matt Muller and Kirsten Perry for PVFleets. Detects low slope periods and generated a clipping threshold for each day with such periods.

## Checklist

### DIFF
--- a/pvanalytics/features/clipping.py
+++ b/pvanalytics/features/clipping.py
@@ -227,10 +227,19 @@ def threshold(ac_power, slope_max=0.0035, power_min=0.75,
     return ac_power >= threshold
 
 
-def geometric(ac_power, window=4, derivative_min=0.2):
-    # TODO downsample and mean if freq of `ac_power` is less than 10 minutes.
+def _downsample(ac_power, freq):
+    if freq is None:
+        freq = pd.infer_freq(ac_power.index)
+    minutes = util.freq_to_timedelta(freq).seconds / 60
+    if minutes > 10:
+        return ac_power
+    return ac_power.resample('15T').mean()
+
+
+def geometric(ac_power, window=4, derivative_min=0.2, freq=None):
     # TODO set larger window for tracking systems
-    ac_power = ac_power.copy()
+    ac_power_original = ac_power.copy()
+    ac_power = _downsample(ac_power, freq)
     daily_min = ac_power.resample('D').transform('max') * 0.1
     ac_power.loc[ac_power < daily_min] = np.nan
     # rolling max/min on reverse ac_power
@@ -241,17 +250,27 @@ def geometric(ac_power, window=4, derivative_min=0.2):
                      / ((rolling_max + rolling_min) / 2) * 100
     clipped = derivative_max < derivative_min
     # TODO expand the clipmask to all points within the window
-    # TODO resample to input freq if downsampled above
-    # TODO calculate a daily clipping threshold and apply to ac_power
-    clipped_power = ac_power[clipped]
-    daily_clipped_max = clipped_power.resample('D').transform('max').reindex(
-        ac_power.index, method='ffill'
-    )
-    daily_clipped_min = clipped_power.resample('D').transform('min').reindex(
-        ac_power.index, method='ffill'
-    )
-    # TODO add +/- 2*ac_power.resample('D').transform(lambda d: d.std())
-    #      to the max and min clipping respectively for data that has been
-    #      downdampled.
-    return (ac_power >= daily_clipped_min) & (ac_power <= daily_clipped_max)
+    if not ac_power.index.equals(ac_power_original.index):
+        # data was down-sampled. reindex to original index. and use mean
+        # to calculate daily clipping threshold.
+        clipped = clipped.reindex(ac_power_original.index, method='ffill')
+        daily_clipped = clipped.resample('D')
+        daily_mean = daily_clipped.transform(
+            lambda day: ac_power_original[day.index][day].mean()
+        )
+        daily_std = daily_clipped.transform(
+            lambda day: ac_power_original[day.index][day].std()
+        )
+        daily_clipped_max = daily_mean + 2 * daily_std
+        daily_clipped_min = daily_mean - 2 * daily_std
+    else:
+        daily_clipped = clipped.resample('D')
+        daily_clipped_max = daily_clipped.transform(
+            lambda day: ac_power_original[day.index][day].max()
+        )
+        daily_clipped_min = daily_clipped.transform(
+            lambda day: ac_power_original[day.index][day].min()
+        )
+    return (ac_power_original >= daily_clipped_min) \
+           & (ac_power_original <= daily_clipped_max)
 
--- a/pvanalytics/features/clipping.py
+++ b/pvanalytics/features/clipping.py
@@ -379,7 +379,7 @@ def geometric(ac_power, window=None, slope_min=0.2, freq=None,
     ac_power = ac_power_original
     freq_minutes = _freq_minutes(ac_power.index, freq)
     if freq_minutes < 10:
-        ac_power = ac_power.resample('15T').mean()
+        ac_power = ac_power.resample('15T', label='right').mean()
     if window is None and tracking and freq_minutes < 30:
         window = 5
     else:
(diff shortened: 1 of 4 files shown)


pvanalytics/features/clipping.py
```diff
@@ -225,3 +225,178 @@ def threshold(ac_power, slope_max=0.0035, power_min=0.75,
...
+    Returns
+    -------
+    minimum : Series
+        `data` transformed to the mean of ``data[mask]`` minus 2 times
+         the standard deviation of ``data[mask]`` on each day.
+    maximum : Series
+        `data` transformed to the mean of ``data[mask]`` plus 2 times
+         the standard deviation of ``data[mask]`` on each day.
+    """
+    daily_mean = _apply_daily_mask(mask, data, pd.Series.mean)
```
cwhanse, line 282: Curious why use the custom function here instead of something like `data[mask].rolling('D').mean()`. Because the index is uneven after `data[mask]`? 

As I understand it, using `lambda` or `apply` is a real performance hit, since those loop in python rather than applying some C code.

pvanalytics/features/clipping.py
```diff
@@ -225,3 +225,178 @@ def threshold(ac_power, slope_max=0.0035, power_min=0.75,
...
+    return clipped_windows
+
+
+def geometric(ac_power, window=None, slope_min=0.2, freq=None,
+              tracking=False):
+    """Identify clipping based on a the shape of the `ac_power`
+    curve on each day.
+
+    Each day is checked for periods where the slope of `ac_power`
+    is low. The power values in these times are used to calculate
```
cwhanse, line 335:
```suggestion
    is small. The power values in these periods are used to calculate
```

pvanalytics/features/clipping.py
```diff
@@ -225,3 +225,178 @@ def threshold(ac_power, slope_max=0.0035, power_min=0.75,
...
+    a minimum and a maximum clipped power level for that day. Any
+    power values that are within this range are flagged as
+    clipped. The methodology for computing the thresholds varies
+    depending on the frequency of `ac_power`. For high frequency
+    data, less than 10 minute timestamp spacing the minimum
```
cwhanse, line 340:
```suggestion
    data (less than 10 minute timestamp spacing) the minimum
```

pvanalytics/features/clipping.py
```diff
@@ -225,3 +225,178 @@ def threshold(ac_power, slope_max=0.0035, power_min=0.75,
...
+    clipped power is the mean of the low-slope period(s) on that
+    day minus 2 times the standard deviation in the same period(s).
+    For higher frequency data the absolute minimum and maximum of
```
cwhanse, line 343:
```suggestion
    For lower frequency data the absolute minimum and maximum of
```

pvanalytics/features/clipping.py
```diff
@@ -225,3 +225,178 @@ def threshold(ac_power, slope_max=0.0035, power_min=0.75,
...
+    Parameters
+    ----------
+    ac_power : Series
+        AC power data.
+    window : int, optional
+        Size of the rolling window used to identify low-slope
+        periods. If not specified and `tracking` is False then
+        `window=3` is used. If not specified and `tracking` is
+        True then `window=5` is used.
+    slope_min : float, default 0.2
```
cwhanse, line 360: Does this have a unit?  AC power / minute?

cwhanse (approved):

pvanalytics/features/clipping.py
wfvining: Yeah, I did it this way so I could use `.transform()` and end up with the same index. The other way is doable, but ends up with a bunch of re-indexing at the end that I seemed error-prone and more difficult to maintain.

I'm not sure about the performance hit from using a `lambda`. I recall hearing the same in the past, but I have not found any recent references to performance loss, I think we would need to do some profiling to see if it really makes a difference here.

cwhanse: OK as is

kanderso-nrel: If I'm the source of the claimed perf loss with `apply`, this is what I was complaining about:

```python
In [23]: import pandas as pd

In [24]: pd.__version__
Out[24]: '0.25.1'

In [25]: s = pd.Series(1, pd.date_range('2019-01-01', freq='h', periods=8760))

In [26]: %time _=s.rolling('d').apply(sum, raw=False)
Wall time: 906 ms

In [27]: %time _=s.rolling('d').apply(sum, raw=True)
Wall time: 36.9 ms

In [30]: %time _=s.rolling('d').sum()
Wall time: 998 µs
```

wfvining: Seems like that is down to the difference between applying the native python `sum()` and `pandas.core.window.Rolling.sum()`?

kanderso-nrel: I agree that comparison is not apples-to-apples, but I'm not immediately finding a way to use the Rolling methods with `apply`.  Do you have a demo snippet you can show?  If there's a way to use `apply` that has timings comparable to those of the rolling methods, I would love to know about it.

wfvining: Good point. You're right that you can't apply those directly in this context.

pvanalytics/features/clipping.py
```diff
@@ -225,3 +225,178 @@ def threshold(ac_power, slope_max=0.0035, power_min=0.75,
...
+
+    Notes
+    -----
+    Based on code from the PVFleets QA project.
+    """
+    ac_power_original = ac_power.copy()
+    ac_power = ac_power_original
+    freq_minutes = _freq_minutes(ac_power.index, freq)
+    if freq_minutes < 10:
+        ac_power = ac_power.resample('15T').mean()
```
wfvining, line 382:
```suggestion
        ac_power = ac_power.resample('15T', label='right').mean()
```

To be consistent with the directionality in the rolling mean/std below.

pvanalytics/features/clipping.py
wfvining: I was thinking this would give the same results, but with improved performance:

```python
def _apply_daily_mask(mask, data, transformation):
    """Apply `f` to the data selected by `mask` on each day.

    Parameters
    ----------
    mask : Series
        Boolean Series with same index as `data`
    data : Series
        Series with the data.
    transformation : str or function
        Any value that can be passed to ``Series.resample().transform()``.

    Returns
    -------
    Series
        Series with same index as `mask` and values assigned by applying
        transformation to data in ``data[mask]`` on each day.
    """
    data = data.copy()
    data[~mask] = np.nan
    return data.resample('D').transform(transformation)
```

We can pass the names of cythonized kernel functions (`'min'`, `'max'`, `'mean'`, and `'std'`) as `transformation` and should get substantially better performance (5ms per call as opposed to 150ms per call). All well and good, but tests fail for 5-minute data.

pvanalytics/features/clipping.py
wfvining: I'm struggling a little to describe this. It is the percent change per window.

cwhanse: Yeah.

```
    derivative_max = ((rolling_max - rolling_min)
                      / ((rolling_max + rolling_min) / 2) * 100)
    clipped = derivative_max < slope_min
```

Suggest "Maximum difference in maximum and minimum power for a window to be flagged as clipped. Units are percent of average power in the interval."

I think this is a maximum not a minimum slope, maybe `slope_max`

pvanalytics/features/clipping.py
wfvining: Using the cythonized functions gives numerically different results due to floating point rounding problems. For example, `tests/features/test_clipping.py::test_geometric_clipping_correct[5T]` ends up with an upper and a lower bound of `62.40000000000005`, while the maximum value in `ac_power` is `62.4` exactly.

cwhanse: weird

wfvining: Can work around this by setting the minimum to be no larger than the absolute maximum candidate clipped value on each day. I'm not completely happy with this as it feels like a bit of a hack, but it does work for the tests I have implemented. It should work in general since the minimum will only ever be set on days where there is data that appears to be clipped. This also should not generally come up in practice; I believe it only arises here because the standard deviation is 0 since there is no noise in the simulated data used for testing. This is possible, but very unlikely in practice.

cwhanse: Or maybe round the returned min, max values to some large number of decimal places, like 8? A comment is appropriate whatever you choose to do.

wfvining: I was starting to think rounding might be the better option. Still end up with one test failure, so I will need to dig in to it a bit more.

wfvining:
@cwhanse I was getting slight variations from the output of Matt's implementation on the test data sets he provided. To address that I went to a left-labeled rolling max/min for the slope calculations (previously I used a right-labeled rolling window since it made the code a little simpler). This shook out some brittleness in the rounding of the min/max clipping thresholds. I went back to the original approach of adjusting the max up if it is too low and the min down if it is too high. Would you take a quick look at `_threshold_mean()` before I merge this?

cwhanse (approved):

PR #95, opened 2020-11-09
DO NOT MERGE - Testing the coverage uploads from github actions

wfvining:
Testing the coverage uploads from github actions with read only token.


wfvining:
Coveralls failure indicates everything is working as expected.

PR #97, opened 2020-11-13
Ignore missing data in clipping.levels()

### LINKED ISSUE #96: Handle missing data gracefully in clipping.levels()
`features.clipping.levels()` raises some very difficult to decipher value errors form numpy when there are NAs in the data. We should either raise a more meaningful and useful exception or simply deal with NAs by dropping them, applying the clipping filter, reindexing, and filling missing values with `False`.

wfvining:
## Checklist

- [x] Closes #96 

### DIFF
--- a/pvanalytics/features/clipping.py
+++ b/pvanalytics/features/clipping.py
@@ -79,6 +79,9 @@ def levels(ac_power, window=4, fraction_in_window=0.75,
     for more information.
 
     """
+    power = ac_power
+    ac_power = ac_power.copy()
+    ac_power.dropna(inplace=True)
     num_bins = np.ceil(1.0 / rtol).astype(int)
     flags = pd.Series(index=ac_power.index, data=False)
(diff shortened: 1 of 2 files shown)


pvanalytics/features/clipping.py
```diff
@@ -79,6 +79,9 @@ def levels(ac_power, window=4, fraction_in_window=0.75,
     for more information.
 
     """
+    power = ac_power
```
cwhanse, line 82: Only editorial, but I think it would be more clear as
```
power = ac_power.copy()
```
operate on `power`, then `reindex_like(ac_power)`

cwhanse (approved):

pvanalytics/features/clipping.py
wfvining: I do like that better.

wfvining:
Coveralls is reporting a coverage change; however, no change is shown in the coverage reports. I'm going to chalk that up to a mistake on the coveralls/CI side.

PR #101, opened 2021-02-11
Detect shadows from fixed objects in GHI data

### LINKED ISSUE #24: Add shadow detection
Should go in the `features` module, I think.

wfvining:
## Description

This is a port of the pvl_detect_shadows function from MATLAB_PVLIB.

## Checklist

- [x] Closes #24

### DIFF
--- a/pvanalytics/features/shading.py
+++ b/pvanalytics/features/shading.py
@@ -81,12 +81,13 @@ def _prepare_images(ghi, clearsky, daytime, interval):
     ghi = ghi[ghi.resample('D').transform('count') == image_width]
     image_times = ghi.index
     ghi_image = _to_image(ghi.to_numpy(), image_width)
-    scaled_ghi = (ghi * 1000) / _smooth(ghi_image).max()
+    scaled_ghi = (ghi * 1000) / np.max(_smooth(ghi_image))
     scaled_clearsky = (clearsky * 1000) / clearsky.max()
     scaled_clearsky = scaled_clearsky.reindex_like(scaled_ghi)
     daytime = daytime.reindex_like(scaled_ghi)
     # Detect clouds.
     clouds = _detect_clouds(scaled_ghi, scaled_clearsky, '50T')
+    cloud_mask = _to_image(clouds.to_numpy(), image_width)
     # Interpolate across days (i.e. along columns) to remove clouds
     # replace clouds with nans
     #
@@ -94,9 +95,9 @@ def _prepare_images(ghi, clearsky, daytime, interval):
     # but the easiest approach is to turn the image into a dataframe and
     # interpolate along the columns.
     cloudless_image = ghi_image.copy()
-    cloudless_image[_to_image(clouds.to_numpy(), image_width)] = np.nan
+    cloudless_image[cloud_mask] = np.nan
     clouds_image = ghi_image.copy()
-    clouds_image[~_to_image(clouds.to_numpy(), image_width)] = np.nan
+    clouds_image[~cloud_mask] = np.nan
     ghi_image = pd.DataFrame(cloudless_image).interpolate(
         axis=0,
         limit_direction='both'
--- a/pvanalytics/features/shading.py
+++ b/pvanalytics/features/shading.py
@@ -307,14 +307,11 @@ def fixed(ghi, daytime, clearsky, interval=None, min_gradient=2):
     gradient = ndimage.morphological_gradient(ghi_boosted, size=(1, 3))
     threshold = gradient > min_gradient  # binary image of wire candidates
 
-    # From here we CAN use skimage we are working with binary
-    # images. We use the grayscale version of these functions rather
-    # than the binary version since the binary version do not support
-    # the 'area_threshold' parameter.
-    three_day_mask = morphology.rectangle(1, 3)
-    wires = morphology.area_opening(
-        morphology.closing(threshold, three_day_mask),
-        area_threshold=200,
+    # From here we CAN use skimage we are working with binary images.
+    three_minute_mask = morphology.rectangle(1, 3)
+    wires = morphology.remove_small_objects(
+        morphology.binary_closing(threshold, three_minute_mask),
+        min_size=200,
         connectivity=2  # all neighbors (including diagonals)
     )
     wires = _clean_wires(wires)
--- a/pvanalytics/features/shading.py
+++ b/pvanalytics/features/shading.py
@@ -309,7 +309,8 @@ def fixed(ghi, daytime, clearsky, interval=None, min_gradient=2):
         interval = util.freq_to_timedelta(
             pd.infer_freq(ghi.index)
         ).seconds // 60
-    # TODO Handle invalid intervals
+    if interval != 1:
+        raise ValueError("Data must be at 1-minute intervals")
     ghi_image, clearsky_image, clouds_image, index = _prepare_images(
         ghi,
         clearsky,
--- a/pvanalytics/features/shading.py
+++ b/pvanalytics/features/shading.py
@@ -265,13 +265,13 @@ def _clean_wires(wires):
 
 
 def fixed(ghi, daytime, clearsky, interval=None, min_gradient=2):
-    """Detects shadows form fixed structures such as wires and poles.
+    """Detects shadows from fixed structures such as wires and poles.
 
     Uses morphological image processing methods to identify shadows
-    from fixed local objects in GHI data. GHI data are are assumed to
+    from fixed local objects in GHI data. GHI data are assumed to
     be reasonably complete with relatively few missing values and at a
     fixed time interval nominally of 1 minute over the course of
-    several months. Shadows are assumed to be relatively short
+    several months. Detection focuses on shadows with relatively short
     duration. The algorithm forms a 2D image of the GHI data by
     arranging time of day along the x-axis and day of year along the
     y-axis. Rapid change in GHI in the x-direction is used to identify
--- a/pvanalytics/features/shading.py
+++ b/pvanalytics/features/shading.py
@@ -191,22 +191,6 @@ def _restore_gaps(wires):
     )
 
 
-def _width(bbox):
-    """Return the width of a bounding box.
-
-    Parameters
-    ----------
-    bbox : tuple
-        4-tuple specifying ``(min_row, min_col, max_row, max_col)``.
-
-    Returns
-    -------
-    int
-        The width of the bounding box in pixels.
-    """
-    return bbox[3] - bbox[1] + 1
-
-
 def _height(bbox):
     """Return the height of a bounding box.
 
--- a/pvanalytics/features/shading.py
+++ b/pvanalytics/features/shading.py
@@ -241,7 +241,22 @@ def _tand(degrees):
 
 
 def _filter_bars(wires, out):
-    """Only keep shadows that are 'bar-shaped' and span multiple days."""
+    """Only keep shadows that are 'bar-shaped' and span multiple days.
+
+    Parameters
+    ----------
+    wires : ndarray
+        Binary image of candidate shadows.
+    out : ndarray
+        An ndarray that will hold the output image resulting from this
+        filtering operation. If None then a new array will be allocated.
+
+    Returns
+    -------
+    ndarray
+        The output array. If `out` is not none, will return `out`.
+    """
+    temp_image = np.ndarray(wires.shape)
     for angle in range(0, 90, 5):
         if angle < 75:
             dim = 20
@@ -258,13 +273,17 @@ def _filter_bars(wires, out):
             rr = mid + np.round((cc - mid) * _tand(90 - angle))
             se[cc, rr.astype('int')] = 1
             se = se.T
-        out = np.logical_or(
+        morphology.binary_opening(wires, se, out=temp_image)
+        np.logical_or(
             out,
-            morphology.binary_opening(wires, se)
+            temp_image,
+            out=out
         )
-        out = np.logical_or(
+        morphology.binary_opening(wires, np.flipud(se), out=temp_image)
+        np.logical_or(
             out,
-            morphology.binary_opening(wires, np.flipud(se))
+            temp_image,
+            out=out
         )
     return out
 
--- a/pvanalytics/features/shading.py
+++ b/pvanalytics/features/shading.py
@@ -57,6 +57,17 @@ def _prepare_images(ghi, clearsky, daytime, interval):
     Performs pre-processing steps on `ghi` and `clearsky` before
     returning images for use in the shadow detection algorithm.
 
+    Parameters
+    ----------
+    ghi : Series
+        Measured GHI.
+    clearsky : Series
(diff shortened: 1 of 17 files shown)


wfvining:
@cwhanse I've pretty much got the Matlab version ported over. This version does not seem to do quite as good a job removing cloud shadows that are "connected" to wires in the demo data from PVLIB_MATLAB. If you have a little time, I would appreciate a second pair of eyes on the `clean_wires` function (or all of it really). Any other feedback on what parameters should be exposed to the user is also very welcome.

Still some work to do on tests and performance. As it stands it is pretty slow, but there are some obvious places for performance improvement (notably in the `clean_wires` function which can be made to operate in place rather than repeatedly copying the image).

pvanalytics/features/shading.py
```diff
@@ -0,0 +1,338 @@
...
+    """
+    return pd.Series(image.flatten(), index=image_index)
+
+
+def _smooth(image):
+    """Smooth `image` by local mean filtering.
+
+    Local means are computed in a 3x3 square surrounding each pixel.
+    """
+    return ndimage.uniform_filter(image, size=15)
```
cwhanse, line 51: "3x3 square surrounding each pixel" like this?

X X X
X O X
X X X

I think `size=15` will be be a 15 x 15 square.  `size=(3, 3)`?

Maybe also `mode='nearest'` to handle windows extending outside the image.

pvanalytics/features/shading.py
```diff
@@ -0,0 +1,338 @@
...
+
+
+def _prepare_images(ghi, clearsky, daytime, interval):
+    """Prepare data as images.
+
+    Performs pre-processing steps on `ghi` and `clearsky` before
+    returning images for use in the shadow detection algorithm.
+
```
cwhanse, line 59: Add brief description of input parameters. It is clear in the usage that e.g. `ghi` and `clearsky` are `Series` with the same index, and `interval` is in minutes

pvanalytics/features/shading.py
```diff
@@ -0,0 +1,338 @@
...
+    clouds_image : np.ndarray
+        Image of the cloudy periods in `ghi`
+    image_times : pandas.DatetimeIndex
+        Index for the data included in the returned images. Leading
+        and trailing days with incomplete data are not included in the
+        image, these times are needed to build a Series from the image
+        later on.
+
+    """
+    # Fill missing days by interpolation. Interpolation is limited to
```
cwhanse, line 75: ?

```suggestion
    # Fill missing times by interpolation. Interpolation is limited to
```

pvanalytics/features/shading.py
```diff
@@ -0,0 +1,338 @@
...
+        limit_direction='both'
+    ).to_numpy()
+    # set night to nan
+    ghi_image[~_to_image(daytime.to_numpy(), image_width)] = np.nan
+    return (
+        ghi_image,
+        _to_image(scaled_clearsky.to_numpy(), image_width),
+        clouds_image,
+        image_times
+    )
```
cwhanse, line 111: This amount of shuffling between Series and 2D formats makes me think it's an opportunity to improve performance.

Maybe save and re-use `cloud_filter = _to_image(clouds.to_numpy(), image_width)`

pvanalytics/features/shading.py
```diff
@@ -0,0 +1,338 @@
...
+
+
+def _detect_clouds(ghi, clearsky_ghi, window='50T'):
+    """Use :py:func:`pvanalytics.clearsky.reno` to detect clouds"""
+    clouds = pvlib.clearsky.detect_clearsky(
+        ghi,
+        clearsky_ghi,
+        ghi.index,
+        window_length=10
```
cwhanse, line 120: `window='50T'` and `window_length=10`? Wondering if `window_length=50` gets what we want directly and avoids `clouds.rolling(window).sum()`

cwhanse: This function is probably affected by [changes](https://github.com/pvlib/pvlib-python/pull/1074) made in pvlib v0.8., where the intervals are changed from left-labeled to centered.

pvanalytics/features/shading.py
```diff
@@ -0,0 +1,338 @@
...
+    out = _fill_gaps(out)
+    out = _restore_gaps(out)
+    out = _filter_blobs(out, 20, connectivity=1)
+    out = _filter_bars(wires, out)
+    out = _filter_blobs(out, 15, connectivity=2)
+    return out
+
+
+def fixed(ghi, daytime, clearsky, interval=None, min_gradient=2):
+    """Detects shadows form fixed structures such as wires and poles.
```
cwhanse, line 267:
```suggestion
    """Detects shadows from fixed structures such as wires and poles.
```

pvanalytics/features/shading.py
```diff
@@ -0,0 +1,338 @@
...
+
+    Uses morphological image processing methods to identify shadows
+    from fixed local objects in GHI data. GHI data are are assumed to
```
cwhanse, line 270:
```suggestion
    from fixed local objects in GHI data. GHI data are assumed to
```

pvanalytics/features/shading.py
```diff
@@ -0,0 +1,338 @@
...
+    be reasonably complete with relatively few missing values and at a
+    fixed time interval nominally of 1 minute over the course of
+    several months. Shadows are assumed to be relatively short
```
cwhanse, line 273:
```suggestion
    several months. Detection focuses on shadows with relatively short
```

pvanalytics/features/shading.py
```diff
@@ -0,0 +1,338 @@
...
+    duration. The algorithm forms a 2D image of the GHI data by
+    arranging time of day along the x-axis and day of year along the
+    y-axis. Rapid change in GHI in the x-direction is used to identify
+    edges of shadows; continuity in the y-direction is used to
+    separate local object shading from cloud shadows.
+
+    Parameters
+    ----------
+    ghi : Series
+        Time series of GHI measurements. Should be at 1-minute frequency
```
cwhanse, line 283: I think we should make 1-minute frequency a requirement. I don't have any evidence that this algorithm works for slightly irregular time spacing, and the conversion to an image assumes that the data have a regular frequency.

pvanalytics/features/shading.py
```diff
@@ -0,0 +1,338 @@
...
+        for best results. Must be 1 year of data.
+    daytime : Series
+        Boolean series with True for times when the sun is up.
+    clearsky : Series
+        Clearsky GHI.
```
cwhanse, line 288:
```suggestion
        Clearsky GHI with the same index as ``ghi``.
```

pvanalytics/features/shading.py
```diff
@@ -0,0 +1,338 @@
...
+        daytime,
+        interval
+    )
+    # normalize the GHI and dampen the dynamic range where the clear
+    # sky model may have large errors (e.g. at very low sun elevation)
+    alpha = 2000
+    ghi_boosted = 1000 * (ghi_image + alpha) / (clearsky_image + alpha)
+
+    # We must use scipy.ndimage here because skimage does not support
+    # floating point the data outside the range [-1, 1]
```
cwhanse, line 324: grammar error in this comment

cwhanse (commented):
It looks well structured and readable. Do we want to add a test that produces an image similar to the output of the matlab code?

pvanalytics/features/shading.py
wfvining: What was I doing?? Thanks. We could use `mode='nearest'`, but I think it wont make much of a difference since the smoothed image is just used to get a maximum ghi value to re-scale the input.

pvanalytics/features/shading.py
wfvining: Might be some room to improve by using scipy interpolation functions as well, avoiding going back to pandas all together. I had a hard time figuring out that API which is why I went back to pandas, do you know how to interpolate along one dimension of an `ndarray`?

pvanalytics/features/shading.py
cwhanse: > I think it wont make much of a difference since the smoothed image is just used to get a maximum ghi value to re-scale the input.

Ok with this, thanks.

pvanalytics/features/shading.py
cwhanse: `ghi_image = pd.DataFrame(` keeps each minute of the day as a column, right?

I'm pretty sure pandas is just wrapping the scipy interpolate functions. Axis-wise interpolation is an [open feature request](https://github.com/scipy/scipy/issues/6187) for scipy.

If the function's performance is not acceptable and this manipulation is the limiting step, we could probably write our own interpolator and skip the pandas and scipy overhead. The result may not be any faster.

OK for now, and we can revisit this point if performance is a problem.

pvanalytics/features/shading.py
wfvining: The way I read the original code is that it used a window of 10 to do the cloud detection, then did a convolution with a 50 minute window to keep only 50-minute or longer cloudy periods. I tried `window_length=50`, but it substantially changes the output.

For pvlib >= 0.8.1 we need `align='left'` I guess. Might be time to bump the minimum version of pvlib to 0.8.1, but that should probably be done separately from this PR since it may require other changes as well.

pvanalytics/features/shading.py
wfvining: Makes my life easier. Should we require any particular number of days, or just sternly warn the users should have several months of data for this to work well?

cwhanse: There's a vertical dimension to the elements used in `_filter_bars`. Maybe the code will run for data with fewer days, but that's not the intent of the algorithm. I'd warn that the algorithm should be applied to at least 60 days of data. If there's a hard lower limit on the vertical dimension of the image, from the `ndimage` functions, we could state that also. May not be worth figuring that limit out.

pvanalytics/features/shading.py
cwhanse: > The way I read the original code is that it used a window of 10 to do the cloud detection, then did a convolution with a 50 minute window to keep only 50-minute or longer cloudy periods.

That sounds right.

>  I tried `window_length=50`, but it substantially changes the output.

Agree. I withdraw my suggestion. Maybe change the parameter to `duration='50T'` to avoid confusion with the `window_length` parameter for the pvlib function.
 
> For pvlib >= 0.8.1 we need `align='left'` I guess. Might be time to bump the minimum version of pvlib to 0.8.1, but that should probably be done separately from this PR since it may require other changes as well.

My comment was made when thinking about matching the Matlab code output. Maybe that's not necessary and this implementation can live with centered intervals for determining clear-sky times.

pvanalytics/features/shading.py
```diff
@@ -0,0 +1,340 @@
...
+            se[cc, rr.astype('int')] = 1
+            se = se.T
+        out = np.logical_or(
+            out,
+            morphology.binary_opening(wires, se)
+        )
+        out = np.logical_or(
+            out,
+            morphology.binary_opening(wires, np.flipud(se))
+        )
```
wfvining, line 250: These two operations are creating 4 copies of the image, we can reduce this by allocating one `ndarray` that `morphology.binary_opening` puts its output into and passing `out=out` to `np.logical_or`. Similar performance improvements can be had in the other helper functions used bye `_clean_wires`.

pvanalytics/features/shading.py
```diff
@@ -0,0 +1,368 @@
...
+
+def _filter_blobs(wires, blob_length, connectivity):
+    """Remove blobs shorter than blob_length.
+
+    Examines all connected components in the image and removes any
+    that span fewer than `blob_length` columns.
+    """
+    components = measure.label(wires, connectivity=connectivity, background=0)
+    keep = [props.label for props in measure.regionprops(components)
+            if _height(props.bbox) > blob_length]
```
wfvining, line 225: Here is one discrepancy from the MATLAB code:

The comments indicate that we should eliminate blobs that span fewer than `blob_length` days; however, in `GHIImage.bwlengthopen` it looks like the width of the bounding box is used instead of the height (`cc` is the connected component):

```matlab
                [~,cols] = ind2sub(size(cc),find(cc == ii));
                len(ii) = max(cols)-min(cols)+1;
```

I found that for the SMUD test data using width instead of height resulted in many clouds being left in the image.

wfvining:
@cwhanse I think this is ready for a closer look. The test failure is in the `features.clearsky.reno()`, not `features.shading`, they will need to be resolved as part of #105. As far as I can tell, with the exception of the comment above, I have faithfully re-implemented the matlab code, but this version is not doing quite as good a job of eliminating the cloud shadows that are 'connected' to wire shadows.

pvanalytics/features/shading.py
wfvining: Matlab images had time of day on the row index, and day of year on the column index. Here we are transposed from that. This doesn't affect the blob detection (since it is already based on `_height()`, but did necessitate a change to the remove pillars step (transposing the masks).

wfvining:
I verified that the structuring elements used in `_filter_bars()` *include* a vertical bar (spanning multiple days) and *exclude* a horizontal bar (within a single day).

pvanalytics/features/shading.py
```diff
@@ -0,0 +1,374 @@
...
+import pvlib
+from pvanalytics import util
+
+
+def _to_image(data, width):
+    """Convert data to an image.
+
+    Parameters
+    ----------
+    data : Series
```
cwhanse, line 15:
```suggestion
    data : array_like
```

pvanalytics/features/shading.py
```diff
@@ -0,0 +1,374 @@
...
+def _prepare_images(ghi, clearsky, daytime, interval):
+    """Prepare data as images.
+
+    Performs pre-processing steps on `ghi` and `clearsky` before
+    returning images for use in the shadow detection algorithm.
+
+    Parameters
+    ----------
+    ghi : Series
+        Measured GHI.
```
cwhanse, line 63:
```suggestion
        Measured GHI. [W/m^2]
```

pvanalytics/features/shading.py
```diff
@@ -0,0 +1,374 @@
...
+    clearsky : Series
+        Expected clearsky GHI.
```
cwhanse, line 65:
```suggestion
        Expected clearsky GHI. [W/m^2]
```

pvanalytics/features/shading.py
```diff
@@ -0,0 +1,374 @@
...
+
+
+def _detect_clouds(ghi, clearsky_ghi, window_size):
+    """Use :py:func:`pvanalytics.clearsky.reno` to detect clouds.
+
+    Returns
+    -------
+    Series
+        Boolean series with true for cloudy periods that are at least as long
+        as `duration`.
```
cwhanse, line 134:
```suggestion
        Boolean series with true for cloudy periods that are at least as long
        as `window_size`.
```

cwhanse (approved):
Nice piece of code @wfvining

PR #108, opened 2021-05-07
updated outliers.zscore to include nan_policy

### LINKED ISSUE #102: Handling missing data in outliers
Hello, first of all very interesting project!

I was trying out the outlier funcions. The `outliers.zscore` function (which depends on `scipy.stats.zscore`) returns a series full of NaNs when at least one NaN is present in data.

Would it be helpful to use the option of ignoring such data? Like `stats.zscore(data, nan_policy='omit')` or dropping NaN values?
Or handling this type of data is not to be contemplated?

MichaelHopwood:
## Description

An update to `outliers.zscore` to expose the nan_policy.

## Checklist

not on this list*

- [x] Closes #102
- [ ] Added new API functions to `docs/api.rst`
- [ ] Clearly documented all new API functions with [PEP257](https://www.python.org/dev/peps/pep-0257/) and [numpydoc](https://numpydoc.readthedocs.io/en/latest/format.html) compliant docstrings
- [ ] Non-API functions clearly documented with docstrings or comments as necessary

### DIFF
--- a/pvanalytics/quality/outliers.py
+++ b/pvanalytics/quality/outliers.py
@@ -71,8 +71,8 @@ def zscore(data, zmax=1.5, nan_policy='raise'):
         elif nan_policy == 'omit':
             nan_mask = data.isna()
         else:
-            raise ValueError("Incorrect specification passed. "
-                             "Expected `raise` or `omit`.")
+            raise ValueError(f"Unnexpected value ({nan_policy}) passed to nan_policy. "
+                             "Expected 'raise' or 'omit'.")
 
     data[~nan_mask] = pd.Series(abs(stats.zscore(data[~nan_mask])) > zmax,
                                 index=data[~nan_mask].index)
--- a/pvanalytics/quality/outliers.py
+++ b/pvanalytics/quality/outliers.py
@@ -61,8 +61,8 @@ def zscore(data, zmax=1.5, nan_policy='raise'):
 
     """
     return pd.Series((abs(stats.zscore(data,
-                                       nan_policy=nan_policy)) > zmax),
-                                       index=data.index)
+                                       nan_policy=nan_policy))
+                      > zmax), index=data.index)
 
 
 def hampel(data, window=5, max_deviation=3.0, scale=None):
--- a/pvanalytics/quality/outliers.py
+++ b/pvanalytics/quality/outliers.py
@@ -71,8 +71,9 @@ def zscore(data, zmax=1.5, nan_policy='raise'):
         elif nan_policy == 'omit':
             nan_mask = data.isna()
         else:
-            raise ValueError(f"Unnexpected value ({nan_policy}) passed to nan_policy. "
-                             "Expected 'raise' or 'omit'.")
+            raise ValueError(f"Unnexpected value ({nan_policy}) "
+                             "passed to nan_policy. Expected "
+                             "'raise' or 'omit'.")
 
     data[~nan_mask] = pd.Series(abs(stats.zscore(data[~nan_mask])) > zmax,
                                 index=data[~nan_mask].index)
--- a/pvanalytics/quality/outliers.py
+++ b/pvanalytics/quality/outliers.py
@@ -71,9 +71,8 @@ def zscore(data, zmax=1.5, nan_policy='raise'):
         elif nan_policy == 'omit':
             nan_mask = data.isna()
         else:
-            raise ValueError(f"Unnexpected value ({nan_policy}) "
-                             "passed to nan_policy. Expected "
-                             "'raise' or 'omit'.")
+            raise ValueError(f"Unnexpected value ({nan_policy}) passed to "
+                             "nan_policy. Expected 'raise' or 'omit'.")
 
     data[~nan_mask] = pd.Series(abs(stats.zscore(data[~nan_mask])) > zmax,
                                 index=data[~nan_mask].index)
--- a/pvanalytics/quality/outliers.py
+++ b/pvanalytics/quality/outliers.py
@@ -71,8 +71,8 @@ def zscore(data, zmax=1.5, nan_policy='raise'):
         elif nan_policy == 'omit':
             nan_mask = data.isna()
         else:
-            raise ValueError("Incorrect specification passed " +
-                             "to zscore's nan_policy.")
+            raise ValueError("Incorrect specification passed. "
+                             "Expected `raise` or `omit`.")
 
     data[~nan_mask] = pd.Series(abs(stats.zscore(data[~nan_mask])) > zmax,
                                 index=data[~nan_mask].index)
--- a/pvanalytics/quality/outliers.py
+++ b/pvanalytics/quality/outliers.py
@@ -38,7 +38,7 @@ def tukey(data, k=1.5):
             | (data > (third_quartile + k*iqr)))
 
 
-def zscore(data, zmax=1.5):
+def zscore(data, zmax=1.5, nan_policy = 'raise'):
     """Identify outliers using the z-score.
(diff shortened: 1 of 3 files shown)


wfvining:
@MichaelHopwood I enabled the CI pipeline for you, sorry about the delay, I think this is a new feature of Github Actions.

MichaelHopwood:
Thanks @wfvining! I believe I fixed the minor linting issues but will need to confirm once the CI pipeline is activated again.

wfvining:
@MichaelHopwood sorry about that... this is a very inconvenient "feature." I'll keep a close eye to try to approve runs quickly. Apparently people were using actions for crypto mining, so there isn't a way to just grant blanked approval.

I would recommend running `pytest pvanalytics` and `flake8` locally to speed up the feedback loop for yourself for the time being.

Once this PR is merged any future contributions should not require maintainer approval for the actions pipeline to run.

MichaelHopwood:
Hi @wfvining, `scipy.stats.zscore` was given the parameter `nan_policy` in release 1.4.0 [1]. For this PR to run successfully, the requirements files would have to reflect the newer version of scipy. I'll leave it up to you to decide. 👍 

[1] http://scipy.github.io/devdocs/release.1.4.0.html#scipy-stats-improvements

wfvining:
Shoot, I did not realize that parameter was added so recently. We've been trying to apply the NEP-29 rules for dependencies, so that means we should not require scipy > 1.2.X yet. 1.4 isn't that new, but still withing 24 months (December, 2019) and 3 minor versions.

It would not be too difficult to mimic the scipy `raise` behavior, the other options look somewhat more complex, and I'm not sure we want to got to the effort of recreating them. Maybe just `raise` is enough, with a useful message that indicates users should get rid of their nans. What do you think?

wfvining:
Another option would be to try passing `nan_policy` and catch the `TypeError` if the older version of scipy is installed. In that case, we would revert to raising our own error if there are nans, or calling `zscore` without the `nan_policy` arg if there are not. We'd probably need to also raise a `ValueError` if the user passes something other than `'raise'` in this case, something that says to upgrade to scipy >= 1.4 if they need that behavior.

cwhanse:
If we're not advancing `scipy` then I would favor mimicking the `raise` and `omit` behavior. I don't think that `propagate` makes much sense here (any `nan` in the input makes the output `nan`). If `raise`, any `nan` in input raises a `ValueError`; if `omit`, the zscore  is returned in position where data is not `nan`, along with `nan` in positions corresponding to input.

wfvining:
`propagate` is the behavior prior to 1.4, so we kind of get that for free. I looked at how scipy handles `omit` and it is a bit more involved than I expected. Maybe for our purposes it wouldn't be so complex: Just drop the nans, compute the zscores, reindex and restore the nans?

cwhanse:
> Just drop the nans, compute the zscores, reindex and restore the nans?

That's what I am thinking.

MichaelHopwood:
Thanks for the feedback! I think I captured the verdict regarding NaN handling.

pvanalytics/quality/outliers.py
```diff
@@ -49,6 +49,11 @@ def zscore(data, zmax=1.5):
         A series of numeric values in which to find outliers.
     zmax : float
         Upper limit of the absolute values of the z-score.
+    nan_policy : str
```
wfvining, line 52:
```suggestion
    nan_policy : str, default 'raise'
```

pvanalytics/quality/outliers.py
```diff
@@ -49,6 +49,11 @@ def zscore(data, zmax=1.5):
...
+        Define how to handle NaNs in the input series.
+        If 'raise', raises a ValueError.
+        If 'omit', removes NaN values, computes zscore,
+        and places False booleans where the NaNs originally were.
```
wfvining, line 56:
```suggestion
        If 'raise', a ValueError is raised when `data` contains NaNs.
        If 'omit', NaNs are ignored and False is returned at indices that
        contained NaN in `data`.
```

pvanalytics/quality/outliers.py
```diff
@@ -57,7 +62,25 @@ def zscore(data, zmax=1.5):
         outlier.
 
     """
-    return pd.Series((abs(stats.zscore(data)) > zmax), index=data.index)
+    data = data.copy()
+    nan_mask = pd.Series([False] * len(data))
+
+    if data.isnull().values.any():
```
wfvining, line 68:
```suggestion
    if data.hasnans:
```

pvanalytics/quality/outliers.py
```diff
@@ -57,7 +62,25 @@ def zscore(data, zmax=1.5):
...
+        if nan_policy == 'raise':
+            raise ValueError("The input contains nan values.")
+        elif nan_policy == 'omit':
+            nan_mask = data.isna()
+        else:
+            raise ValueError("Incorrect specification passed " +
+                             "to zscore's nan_policy.")
```
wfvining, line 75: Add the valid options to the error message, and remove the name of the function (it will already be visible in the stack trace).

Also, don't need `+` for strings, python will concatenate adjacent strings automatically in this context:
```suggestion
            raise ValueError("Incorrect specification passed "
                             "to zscore's nan_policy.")
```

pvanalytics/tests/quality/test_outliers.py
```diff
@@ -43,6 +43,25 @@ def test_tukey_lower_criteria():
...
 
+def test_zscore_raise_nan_input():
+    data = pd.Series([1, 0, -1, 0, np.NaN, 1, -1, 10])
+
+    try:
+        outliers.zscore(data, nan_policy='raise')
+    except ValueError:
+        assert True
+    else:
+        assert False
```
wfvining, line 54: Take a look at [`pytest.raises()`](https://docs.pytest.org/en/6.2.x/assert.html#assertions-about-expected-exceptions)

pvanalytics/quality/outliers.py
wfvining: > Add the valid options to the error message, and remove the name of the function.

Can you add this as well? It really makes a difference for users to be able to address errors quickly.

MichaelHopwood: Incorporated! Thanks!

pvanalytics/quality/outliers.py
```diff
@@ -57,7 +62,25 @@ def zscore(data, zmax=1.5):
...
+    nan_mask = pd.Series([False] * len(data))
+
+    if data.hasnans:
+        if nan_policy == 'raise':
+            raise ValueError("The input contains nan values.")
+        elif nan_policy == 'omit':
+            nan_mask = data.isna()
+        else:
+            raise ValueError("Incorrect specification passed. "
+                             "Expected `raise` or `omit`.")
```
wfvining, line 75:
```suggestion
            raise ValueError(f"Unnexpected value ({nan_policy}) passed to nan_policy. "
                             "Expected 'raise' or 'omit'.")
```

pvanalytics/tests/quality/test_outliers.py
cwhanse: @MichaelHopwood suggest

```
with pytest.raises(ValueError, match='Incorrect specification passed'):
    outliers.zscore(data, nan_policy='raise')
```

cwhanse (approved):

wfvining (commented):
This is my last nit to pick on error messages, I swear.

wfvining:
Just needs coverage for the unnexpected nan_policy value. Then good-to-go.

wfvining (approved):

PR #111, opened 2021-11-15
Add convenience function for calculating benchmarking metrics

AdamRJensen:
## Description

There exists a number of metrics that can be used for comparing a modeled time-series against a measured (reference) time-series. I propose to add a convenience function that makes it easy to calculate the most commonly used performance/benchmarking metrics. An overviewof benchmarking metrics for solar irradiance time-series was made by [Gueymard (2014)](https://doi.org/10.1016/j.rser.2014.07.117).

## Checklist

not on this list*

- [ ] Closes #xxx
- [ ] Added new API functions to `docs/api.rst`
- [ ] Clearly documented all new API functions with [PEP257](https://www.python.org/dev/peps/pep-0257/) and [numpydoc](https://numpydoc.readthedocs.io/en/latest/format.html) compliant docstrings
- [ ] Non-API functions clearly documented with docstrings or comments as necessary
- [ ] Added tests to cover all new or modified code
- [ ] Pull request is nearly complete and ready for detailed review

### DIFF
--- a/pvanalytics/metrics.py
+++ b/pvanalytics/metrics.py
@@ -2,6 +2,10 @@
 
 from pvlib.temperature import sapm_cell
 from pvlib.pvsystem import pvwatts_dc
+from dataclasses import dataclass, field
+from typing import Union
+import numpy as np
+import pandas as pd
 
 
 def _performance_ratio(measured, modeled):
@@ -115,3 +119,58 @@ def variability_index(measured, clearsky, freq=None):
        2012.
     """
     return _calc_pathlength(measured, freq) / _calc_pathlength(clearsky, freq)
+
+
+@dataclass
+class BenchmarkingMetrics:
+
+    # Define inputs and their possible types
+    measured: Union[pd.Series, np.array] = field(default=None)
+    modeled: Union[pd.Series, np.array] = field(default=None)
+
+    def __post_init__(self):
+        self.N = len(self.measured)
+
+    # Absolute Mean Bias Deviation (aMBD).
+    def MBD(self):
+        return (self.modeled - self.measured).sum()
+
+    # Relative Mean Bias Deviation (rMBD)
+    def rMBD(self):
+        return (self.modeled - self.measured).sum() / self.measured.mean()
+
+    # Absolute Root Mean Square Deviation (aRMSD)
+    def RMSD(self):
+        return np.sqrt(((self.modeled - self.measured)**2).sum() / self.N)
+
+    # Relative Root Mean Square Deviation (rRMSD)
+    def rRMSD(self):
+        return (np.sqrt(((self.modeled - self.measured)**2).sum() / self.N)
+                / self.measured.mean())
+
+    # Absolute Mean Absolute Deviation (aMAD)
+    def aMAD(self):
+        return np.abs(self.modeled - self.measured).sum()
+
+    # Absolute Mean Absolute Deviation (aMAD)
+    def rMAD(self):
+        return np.abs(self.modeled-self.measured).sum() / self.measured.mean()
+
+    # Absolute Standard Deviation (SD)
+    # By default Pandas uses ddof=1 and Numpy uses ddof=0
+    def aSD(self):
+        return np.sqrt(
+            (self.N*(self.modeled-self.measured)**2).sum()
+            -((self.modeled-self.measured).sum())**2)/self.N
+
+    ## Relative Standard Deviation (SD)
+    def rSD(self):
+        return self.aSD() / self.measured.mean()
+
+    # Coefficient of determination (R^2)
+    def R_squared(self):
+        return (
+            ((self.modeled-self.modeled.mean()) *
+             (self.measured-self.measured.mean())).sum() /
+            ((self.modeled-self.modeled.mean())**2 *
+             (self.measured-self.measured.mean())**2))**2


AdamRJensen:
@cwhanse As I've never worked with data classes before, I'd appreciate it if you just gave a thumbs up if you think I'm on the right path with what I've written so far.

kanderso-nrel:
> @functools.cached_property

fyi I think this only exists in python >=3.8

AdamRJensen:
Thanks for the feedback @wfvining. I don't have a personal preference, and probably lack sufficient experience to determine which of the two options are preferred. Perhaps @cwhanse or @kanderso-nrel has an opinion?

cwhanse:
I'm leery of the deferred computation. I think it's confusing for less-experienced users, and can lead to difficulty re-using components from this library. I'd prefer the `DataClass` with only attributes, and the `compare_series` function.

wholmgren:
Is there going to be any consideration for resampling, interval convention, or filtering? Or is that left to the user? My own experience is these trickier aspects are most relevant when formalizing in a class. But that's a much larger scope than what's currently proposed.

AdamRJensen:
@wholmgren My initial thought was to let the user handle this. I don't really see how this could look... perhaps you can elaborate?

In regards to interval convention, then I suppose it doesn't matter much, assuming that the modeled and measured time-series follow the same convention.

I would propose that a simple nan check is implemented, e.g., an error is raised if either time-series contains nans.

wholmgren:
> My initial thought was to let the user handle this... In regards to interval convention, then I suppose it doesn't matter much, assuming that the modeled and measured time-series follow the same convention.

Totally reasonable to let the user handle it. You seem like the kind of person that is very careful with comparisons so maybe this is of relatively little value to you. It's clear to me that many others are not so careful. Also here's some relevant [discussion](https://groups.google.com/g/pvlib-python/c/b4uMI3v3qOc/m/TIDt9FcnAQAJ) in the pvlib google group.

>  I don't really see how this could look... perhaps you can elaborate?

I'm not saying this is the right way to do it here, but (no surprise) I'd point to [`solarforecastarbiter`](https://github.com/SolarArbiter/solarforecastarbiter-core) for a highly formal example. And obviously the forecast-specific attributes should be ignored in this discussion. Workflow goes like this:

1. Define an [`Observation`](https://github.com/SolarArbiter/solarforecastarbiter-core/blob/262d4a2c3df8b2388fd4a0196bb0cbf810bc45f2/solarforecastarbiter/datamodel.py#L517-L567) and a [`Forecast`](https://github.com/SolarArbiter/solarforecastarbiter-core/blob/262d4a2c3df8b2388fd4a0196bb0cbf810bc45f2/solarforecastarbiter/datamodel.py#L730-L802). These include the interval conventions. 
2. Put them together in a [`ForecastObservation`](https://github.com/SolarArbiter/solarforecastarbiter-core/blob/262d4a2c3df8b2388fd4a0196bb0cbf810bc45f2/solarforecastarbiter/datamodel.py#L1292-L1329). This step helps to define metadata for additional processing steps.
3. Pass the `ForecastObservation` to a function, [`process_forecast_observations`](https://github.com/SolarArbiter/solarforecastarbiter-core/blob/262d4a2c3df8b2388fd4a0196bb0cbf810bc45f2/solarforecastarbiter/metrics/preprocessing.py#L650-L652) that performs the "preprocessing" steps of making interval labels and lengths consistent, filters data, and drops or fills missing data. 

What you get from this is a [`ProcessedForecastObservation`](https://github.com/SolarArbiter/solarforecastarbiter-core/blob/262d4a2c3df8b2388fd4a0196bb0cbf810bc45f2/solarforecastarbiter/datamodel.py#L1602-L1651) that contains resampled, aligned, filtered time series.

The final step is to pass that object to a function that calculates metrics. Roughly speaking, this PR is a different take on that step. I'd be happy to say this PR should be restricted to that step and set aside the rest of this discussion. I just wanted to bring up the larger context.

> I would propose that a simple nan check is implemented, e.g., an error is raised if either time-series contains nans.

Ah, I suspect there's a variety of opinions here. Pandas default style is to ignore nans. Numpy returns nans unless using the `nan*` functions. I can see an argument for raising too. If you dig into solar forecast arbiter preprocessing.py you'll find there's a sophisticated reporting system for various data checks, including nans, nighttime, unphysical values, etc.

pvanalytics started with a port of the general purpose validation code developed under solarforecastarbiter. I think it would be useful to see a general purpose port of the time series resample/align/fill code that continues to integrate with validation code. But again, maybe a different PR and for now we should only keep it in mind as we think about interfaces.

Also, you might consider adapting the tests in [`test_deterministic`](https://github.com/SolarArbiter/solarforecastarbiter-core/blob/master/solarforecastarbiter/metrics/tests/test_deterministic.py) to this PR. There are a few edge cases to consider.

cwhanse:
Thanks very much @wholmgren.

> resampled, aligned, filtered time series

I'm assuming that those are the inputs to the metrics function proposed here.

> I think it would be useful to see a general purpose port of the time series resample/align/fill code 

I agree.

wholmgren:
> I'm assuming that those are the inputs to the metrics function proposed here.

Correct. I eventually talked myself into believing that we should stay focused on metrics in this PR, but with an acknowledgement that there's much more to consider.

cwhanse:
@AdamRJensen add this line to `requirements-min.txt` file:

`dataclasses`

@wfvining we need to change the last line of the `README`:

```
* `dataclasses` contains classes for normalizing data (e.g. an
  `IVCurve` class)
```

cwhanse:
Test failure being fixed in #115

AdamRJensen:
Should the function be able to handle nans in the two time series?

For example, start by dropping the rows where there is a nan in either the measured or modeled time series?

cwhanse:
> Should the function be able to handle nans in the two time series?

Yes, I think default behavior should be to ignore nans. I'm on the fence if that behavior should be controlled by a kwarg.

PR #118, opened 2021-12-14
Update pandas requirements

### LINKED ISSUE #82: Support pandas >= 1.1.0
Pandas seems to have ironed out the `.groupby().rolling()` bugs that were introduced in version 1.1.0. Unfortunately, while the two obvious bugs that were initially holding up this change have been fixed, the behavior seems to have changed. In particular these two tests fail at pandas 1.1.3:

- `tests/features/test_daytime.py::test_daytime_split_day`
- `tests/features/test_daytime.py::test_daytime_daylight_savings`

The output for both classifies some data that is obviously daytime as night. Might still be a pandas bug, but I'm not sure at this point.

kanderso-nrel:
## Description

This PR updates the requirements spec to prohibit `pandas==1.1.*`.  I figured there was little point in cooking up a workaround that works on all pandas versions because there were only ~6 months between 1.0.5 and 1.2.0 and it's been about a year since 1.2.0 was released.

I tested a range of pandas versions locally (windows) before opening this PR and include the results for reference:
- `pandas==1.1.0`: access violation/segfault
- `pandas>=1.1.1,<=1.1.5`: test failure
- `pandas>=1.2.0`: pass

## Checklist

- [x] Closes #82
- ~Added new API functions to `docs/api.rst`~
- ~Clearly documented all new API functions with [PEP257](https://www.python.org/dev/peps/pep-0257/) and [numpydoc](https://numpydoc.readthedocs.io/en/latest/format.html) compliant docstrings~
- ~Non-API functions clearly documented with docstrings or comments as necessary~
- ~Added tests to cover all new or modified code~

### DIFF
--- a/.github/workflows/lint-and-test.yml
+++ b/.github/workflows/lint-and-test.yml
@@ -7,7 +7,7 @@ jobs:
 
     strategy:
       matrix:
-        os: [ubuntu-latest, macos-latest, windows-latest]
+        os: [ubuntu-latest, macos-10.15, windows-latest]  # macos-latest does not have py3.6
         python-version: [3.6, 3.7, 3.8]
         requirements: [requirements.txt]
         include:
--- a/docs/whatsnew/0.1.1.rst
+++ b/docs/whatsnew/0.1.1.rst
@@ -20,7 +20,10 @@ Bug Fixes
 ~~~~~~~~~
 
 * Added nan_policy to zscore calculation
-  :py:func:`pvanalytics.quality.outliers.zscore`.
+  :py:func:`pvanalytics.quality.outliers.zscore`. (:issue:`102`, :pull:`108`)
+* Prohibit pandas versions in the 1.1.x series to avoid an issue in
+  ``.groupby().rolling()``.  Newer versions starting in 1.2.0 and older
+  versions going back to 0.24.0 are still allowed. (:issue:`82`, :pull:`118`)
 
 Contributors
 ~~~~~~~~~~~~
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
 numpy>=1.15.0
-pandas>=0.24.0,<1.1.0
+pandas>=0.24.0,!=1.1.*
 pvlib>=0.8.0
 scipy>=1.2.0
 statsmodels>=0.9.0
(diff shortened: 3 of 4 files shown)


.github/workflows/lint-and-test.yml
```diff
@@ -7,7 +7,7 @@ jobs:
 
     strategy:
       matrix:
-        os: [ubuntu-latest, macos-latest, windows-latest]
+        os: [ubuntu-latest, macos-10.15, windows-latest]  # macos-latest does not have py3.6
```
kanderso-nrel, line 10: This has nothing to do with the pandas change; it's because `macos-latest` recently changed to macos 11 instead of macos 10, and they dropped python 3.6 in macos 11.  That made the 3.6 test fail here because this is the first PR since the change.  See the table here: https://github.com/actions/virtual-environments/issues/4060.

We might consider dropping python 3.6 from testing though as it is slated to reach EOL in a week or two: https://devguide.python.org/#status-of-python-branches

cwhanse: +1 to dropping support for python 3.6. I'm OK to do that in this PR if it simplifies the pandas requirement.

kanderso-nrel: I don't think dropping 3.6 changes anything for the pandas requirement.  pvanalytics's current minimum (0.24.0) supports python 3.7, so dropping 3.6 is not itself an excuse to bump our minimum and skip over the problematic 1.1.x series.

wfvining:
Do you think we want to prohibit pandas 1.1 across the board? It's just a few functions that depend on the rolling groupby functionality. I'd kind of like to just add a guard in front of those functions that causes an error if the 1.1.x is installed (or issues a warning), but I'm not sure of the added maintenance burden on that approach.

kanderso-nrel:
Normally I'd also support doing a little gymnastics to not drop support unnecessarily, so I wouldn't object to some kind of function-level version check here.  RdTools has its own set of pandas restrictions ([link](https://github.com/NREL/rdtools/blob/master/setup.py#L41-L42)), so that's another motivator, at least to the extent that we want pvanalytics and RdTools to be used in tandem.

cwhanse:
I say we merge this for v0.2 and add an issue to drop support for py36 and pandas<1.2 in v0.3.

cwhanse:
Thanks @kanderso-nrel

Issue #119, opened 2022-01-04
Linear Algebra Error thrown in earlier numpy versions when running orientation fixed_nrel() function.


kperrynrel:
Following error is thrown when running fixed_nrel() function on a azimuth/tilt estimation pipeline:

raise LinAlgError("SVD did not converge in Linear Least Squares")

LinAlgError: SVD did not converge in Linear Least Squares

This occurred when running numpy=1.19.1. Updated to numpy 1.19.5 and issue was resolved. Need to update minimum numpy installation.

cwhanse:
Ah, nvm I see the issue. [This helper function](https://github.com/pvlib/pvanalytics/blob/bb451a030517e0ef180a02d20484261c0897f278/pvanalytics/util/_fit.py#L6) uses `np.polyfit` which is the old numpy API, not yet deprecated AFAIK but less well-mannered than its replaced `np.polynomial`.  We should bring the helper function up to the newer API; transition guide [here](https://numpy.org/doc/stable/reference/routines.polynomials.html).

kperrynrel:
I can do that @cwhanse, will cut a PR making an update on that function.

kanderso-nrel:
Closed by #120?

kperrynrel:
@kanderso-nrel  correct. Closing this--thanks for tracking housekeeping on this!

PR #120, opened 2022-01-04
Updated the polyfit function to np.polynomial.Polynomial.fit()

kperrynrel:
## Description

Addresses issue #119, by updating the np.polyfit() function to newer np.polynomial.Polynomial.fit() function.

### DIFF
--- a/pvanalytics/util/_fit.py
+++ b/pvanalytics/util/_fit.py
@@ -5,7 +5,7 @@ import scipy.optimize
 
 def _quadratic(xs, ys):
     # fit a quadratic function of `xs` to the data in `ys`
-    coefficients = np.polyfit(xs, ys, 2)
+    coefficients = np.polynomial.Polynomial.fit(xs, ys, 2)
     return np.poly1d(coefficients)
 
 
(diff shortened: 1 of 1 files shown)


pvanalytics/util/_fit.py
```diff
@@ -4,9 +4,10 @@
 
 
 def _quadratic(xs, ys):
-    # fit a quadratic function of `xs` to the data in `ys`
-    coefficients = np.polyfit(xs, ys, 2)
-    return np.poly1d(coefficients)
+    # fit a quadratic function of `xs` to the data in `ys`  
+    coefficients = list(np.polynomial.Polynomial.fit(xs, ys, 2).convert())
```
cwhanse, line 8: I think if you use `polynomial.polyfit` and `flipud` this would be simpler.

kperrynrel: image

Tried it out and verified same results--thank you for the suggestion @cwhanse !

cwhanse (approved):

kperrynrel:
@cwhanse what's our policy here on merging PR's in this package, ie are there a certain amount of reviewers I need to check off on this? Or are we good to merge and close now?

cwhanse:
@kperrynrel was just thinking about that. If either Will or I approve and we don't indicate that the other needs to look, then it's OK to merge. If your work is held up by a PR, feel free to nag us about it.

kperrynrel:
Awesome, thanks for the clarification @cwhanse !

PR #124, opened 2022-01-19
Data shift algorithm

kperrynrel:
## Description

This PR handles detecting data shifts in daily summed time series (power or irradiance), and removing data shift periods. The process of doing this is illustrated in the Jupyter notebook in the /docs/ folder.  I am reporting the results of this algorithm (performance) in a submitted PVSC abstract/paper, which we can cite for performance metrics.

## Checklist

not on this list*

- [X] Added new API functions to `docs/api.rst`
- [X] Clearly documented all new API functions with [PEP257](https://www.python.org/dev/peps/pep-0257/) and [numpydoc](https://numpydoc.readthedocs.io/en/latest/format.html) compliant docstrings
- [X] Adds description and name entries in the appropriate "what's new" file 
- [X] Non-API functions clearly documented with docstrings or comments as necessary
- [X] Added tests to cover all new or modified code
- [X] Pull request is nearly complete and ready for detailed review


pvanalytics/tests/quality/test_data_shifts.py
```diff
@@ -0,0 +1,82 @@
...
+import pandas as pd
+import ruptures as rpt
+import pytest
+from pvanalytics.quality import data_shifts as dt
+
+
+@pytest.fixture
+def generate_daily_time_series():
+    # Pull down the saved PVLib dataframe and process it
+    df = pd.read_csv("https://datahub.duramat.org/dataset/7b72ae24-c0c2-4339"
```
kanderso-nrel, line 11: I wonder if we should bundle this file with the package instead of downloading it?  It's 142 KiB, so not huge.  I imagine future tests might want their own datasets, so maybe we should add a `pvanalytics/data` folder like pvlib has.

kanderso-nrel (commented):
Here's a partial review -- I will take another look after some of the higher level questions below are discussed.

Other thoughts:
- Probably makes sense to add a bullet in `docs/index.rst` under `Library Overview` for this new `quality.data_shifts` submodule
- `quality/__init__.py` needs an entry for `data_shifts`

kperrynrel:
@kanderso-nrel I added a /data/ folder containing those two example time series, like you suggested (they're the same as what's on DuraMAT). Is there a good way for referencing the file path when running unit tests? Just worried the scripts aren't going to find them--for pvlib, do you guys reference specific files as test cases for unit tests + jupyter documentation?

kanderso-nrel:
pvlib's approach, which works well enough IMHO, is to find the location of the data directory in relation to the test file rather than via some absolute path.  For pvlib, since many test modules want to access the data directory, we do this centrally in conftest.py and then import it into the test files:
- Define the `DATA_DIR` variable in conftest.py ([example](https://github.com/pvlib/pvlib-python/blob/d5d1d66aae4913f0e23b9a79c655efa1bdafe5f4/pvlib/tests/conftest.py#L72-L73))
- Import it in the test file and get the path to specific files of interest ([example](https://github.com/pvlib/pvlib-python/blob/d5d1d66aae4913f0e23b9a79c655efa1bdafe5f4/pvlib/tests/iotools/test_sodapro.py#L11-L17))
- Pass the filepath to `pd.read_csv` or whatever ([example](https://github.com/pvlib/pvlib-python/blob/d5d1d66aae4913f0e23b9a79c655efa1bdafe5f4/pvlib/tests/iotools/test_sodapro.py#L188))

We have a similar approach for using those data files in the docs (see this [gallery example](https://pvlib-python.readthedocs.io/en/latest/gallery/soiling/plot_greensboro_kimber_soiling.html)), although we should probably improve on that at some point (https://github.com/pvlib/pvlib-python/issues/924).

pvanalytics/quality/data_shifts.py
```diff
@@ -0,0 +1,306 @@
+"""
+Quality tests related to detecting and filtering out data shifts in data
+streams.
+"""
+
+from pvanalytics.quality import gaps
+import numpy as np
+import pandas as pd
+import ruptures as rpt
```
cwhanse, line 9: I think `ruptures` should be an optional dependency, at least for now.

pvanalytics/quality/data_shifts.py
```diff
@@ -0,0 +1,306 @@
...
+    # Run the filtering sequence, if marked as True
+    if filtering:
+        time_series = _erroneous_filter(time_series)
+    # Check if the time series is more than 2 years long. If so, remove
+    # seasonality. If not, run analysis on the normalized time series
+    if (time_series.index.max() - time_series.index.min()).days <= 730:
+        warnings.warn("The passed time series is less than 2 years in length, "
+                      "and cannot be corrected for seasonality. Runnning data "
+                      "shift detection on the min-max normalized time series "
+                      "(NO seasonality correction).")
```
cwhanse, line 195:
```suggestion
                      "and will not be corrected for seasonality. Runnning data "
                      "shift detection on the min-max normalized time series "
                      "with no seasonality correction.")
```

pvanalytics/quality/data_shifts.py
```diff
@@ -0,0 +1,306 @@
...
+        Time series, after data processing. This includes min-max
+        normalization, and, if the time series is in greater than 2 years in
+        length, seasonality removal.
+    """
+    # Convert the time series to a dataframe to do the pre-processing
+    column_name = time_series.name
+    if column_name is None:
+        column_name = "value"
+        time_series = time_series.rename(column_name)
+    df = time_series.to_frame()
```
cwhanse, line 118: Because the input is a single Series, and the constructed series is returned, is the `name` construction only so that the returned `Series.name` has a value? If that's not important, then the constructed Series name is only internal here and could be anything.

pvanalytics/quality/data_shifts.py
```diff
@@ -0,0 +1,306 @@
...
+        series. If True, then this data is filtered out before running the
+        data shift detection sequence. If False, this data is not filtered
+        out. Default set to True.
+    use_default_models: Boolean.
+        If set to True, then default CPD model parameters are used, based
+        on the length of the time series (Window-based models for time series
+        shorter than 2 years in length and BottomUp models for time series
+        longer than 2 years in length). If set to True, none of the method +
+        cost + penalty variables are used.
+    method: ruptures search method object.
```
cwhanse, line 163: I would prefer default to None. Same for `cost`.
```suggestion
    method: ruptures search method instance or None, default None.
```

pvanalytics/quality/data_shifts.py
```diff
@@ -0,0 +1,306 @@
...
+        Ruptures method object.See the following documentation for further
+        information:
+        https://centre-borelli.github.io/ruptures-docs/user-guide/. Default set
+        to ruptures.BottomUp.
+    cost: str
```
cwhanse, line 168:
```suggestion
    cost: str or None, default None
```

pvanalytics/quality/data_shifts.py
```diff
@@ -0,0 +1,306 @@
...
+        If set to True, then default CPD model parameters are used, based
+        on the length of the time series (Window-based models for time series
+        shorter than 2 years in length and BottomUp models for time series
+        longer than 2 years in length). If set to True, none of the method +
+        cost + penalty variables are used.
+    method: ruptures search method object.
+        Ruptures method object.See the following documentation for further
+        information:
+        https://centre-borelli.github.io/ruptures-docs/user-guide/. Default set
+        to ruptures.BottomUp.
```
cwhanse, line 167:
```suggestion
        Ruptures search method instance. See 
        https://centre-borelli.github.io/ruptures-docs/user-guide/.
```

pvanalytics/quality/data_shifts.py
```diff
@@ -0,0 +1,306 @@
...
+    cost: str
+        Cost function passed to the ruptures changepoint detection method.
+        See the following documentation for further
+        information: https://centre-borelli.github.io/ruptures-docs/user-guide/
+        Default set to "rbf".
```
cwhanse, line 172:
```suggestion
        Cost function passed to the ruptures changepoint search instance.
        See https://centre-borelli.github.io/ruptures-docs/user-guide/
```

pvanalytics/quality/data_shifts.py
```diff
@@ -0,0 +1,306 @@
...
+        Whether or not to filter out outliers and stale data from the time
+        series. If True, then this data is filtered out before running the
+        data shift detection sequence. If False, this data is not filtered
+        out. Default set to True.
+    use_default_models: Boolean.
+        If set to True, then default CPD model parameters are used, based
+        on the length of the time series (Window-based models for time series
+        shorter than 2 years in length and BottomUp models for time series
+        longer than 2 years in length). If set to True, none of the method +
+        cost + penalty variables are used.
```
cwhanse, line 162: I'm OK with this boolean parameter, to relieve a user of needing to understand `method` and `cost`. I would prefer to be a bit more specific in the text about what `method` and `cost` are assumed if `use_default` is True
```suggestion
        If True, then default change point detection search parameters are used.
        For time series shorter than 2 years in length, the search function is `rpt.Window`
        with `model='rbf'`, `width=40` and `penalty=30`. For time series 2 years or
        longer in length, the search function is `rpt.BottomUp` with `model='rbf'`
        and `penalty=40`.
```

pvanalytics/quality/data_shifts.py
```diff
@@ -0,0 +1,306 @@
...
+    else:
+        algo = method(model=cost).fit(points)
+        result = algo.predict(pen=penalty)
+    # Remove the last index of the time series, if present
+    if len(points) in result:
+        result.remove(len(points))
+    # Return a list of dates where changepoints are detected
+    time_series_processed.index.name = "datetime"
+    time_series_processed = time_series_processed.reset_index()
+    return list(time_series_processed.loc[result]['datetime'])
```
cwhanse, line 227: I would prefer that this returned `Series(index=time_series.index)` of Boolean, True where `result` is True.

pvanalytics/quality/data_shifts.py
```diff
@@ -0,0 +1,335 @@
...
+
+
+def filter_data_shifts(time_series, filtering=True, use_default_models=True,
```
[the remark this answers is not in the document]
cwhanse: If `detect_data_shifts` returns a Series of boolean, then this function can return the (start, end) pairs defining intervals of change points, or, just the indices of the change points. I see value in that.  I agree that `filter_` implies that it would return a subset of the input Series.

pvanalytics/tests/quality/test_data_shifts.py
cwhanse: +1 to including the file and to setting up `pvanalytics/data`.

cwhanse (commented):
Mostly docstring and API comments. I can look at the tests in another round.

kanderso-nrel (commented):
I think the notebook is a bit out of date (function rename), but I lean towards splitting that into its own PR anyway, so maybe just remove it here and we'll add it once #127 is merged?

cwhanse:
@kperrynrel I can review but prefer to wait until after @kanderso-nrel's comments are addressed (I agree with all).

kanderso-nrel:
FYI the test failures should be fixed by the last commit in #129

kperrynrel:
@cwhanse I've got the issues that @kanderso-nrel brought up addressed, so this should be ready for review. Everything passing, but our coveralls coverage dropped by .2%.

PR #127, opened 2022-01-26
Set up sphinx gallery; add clearsky.reno example

kanderso-nrel:
## Description

This PR:
- Sets up an example gallery using `sphinx-gallery`: [link](https://pvanalytics--127.org.readthedocs.build/en/127/generated/gallery/index.html)
- Adds a short example of `clearsky.reno`: [link](https://pvanalytics--127.org.readthedocs.build/en/127/generated/gallery/clearsky-detection.html#sphx-glr-generated-gallery-clearsky-detection-py)
- Configures function-level docs pages to have a "mini-gallery" at the bottom listing examples using that function: [link](https://pvanalytics--127.org.readthedocs.build/en/127/generated/pvanalytics.features.clearsky.reno.html#examples-using-pvanalytics-features-clearsky-reno)

## Checklist

- [x] Closes #125
- ~Added new API functions to `docs/api.rst`~
- ~Clearly documented all new API functions with [PEP257](https://www.python.org/dev/peps/pep-0257/) and [numpydoc](https://numpydoc.readthedocs.io/en/latest/format.html) compliant docstrings~
- [x] Adds description and name entries in the appropriate "what's new" file 
      in [`docs/whatsnew`](https://github.com/pvlib/pvanalytics/tree/master/docs/whatsnew) 
      for all changes. Includes link to the GitHub Issue with `` :issue:`num` `` 
      or this Pull Request with `` :pull:`num` ``. Includes contributor name 
      and/or GitHub username (link with `` :ghuser:`user` ``).
- ~Non-API functions clearly documented with docstrings or comments as necessary~
- ~Added tests to cover all new or modified code~
- [x] Pull request is nearly complete and ready for detailed review

PR #128, opened 2022-01-27
Get clearsky.reno tests passing again

### LINKED ISSUE #105: Tests for clearsky.reno fail with pvlib 0.8.1
PVLib 0.8.1 changed the labeling of clear-sky periods (https://github.com/pvlib/pvlib-python/pull/1074) which is resulting in a test failures in `test_clearsky.py`.

The change in pvlib labels windows on the center by default, while previous behavior labeled on the left. We probably want to accept the new default behavior from PVLib (updating the tests to match) and expose the new `align` parameter in `pvanalytics.clearsky.reno`.

### LINKED ISSUE #126: Error in clearsky.reno in newer pandas and numpy versions
I know there are some known pandas issues already, but I think this one is new?  This snippet raises an error for some combinations of pandas and numpy:
```python
from pvanalytics.features.clearsky import reno
import pandas as pd

times = pd.date_range('2019-01-01 12:00', freq='1min', periods=3)
ghi = pd.Series([1,1,1], index=times)
reno(ghi, ghi)  # ValueError: clearsky requires regular time intervals of 15m or less
```

It fails with `numpy==1.21.4` + `pandas==1.3.0`.  It works with `numpy==1.19.5` + `pandas==1.2.5`.

This is the issue:

```python
In [7]: pd.Timedelta("1T") / np.timedelta64(1, "60s")
Out[7]: 60.0  # returns 1.0 on older versions
```

I did not do any digging for changes in either library that may have caused this, or whether the issue can be isolated to a change in just one of pandas or numpy.

kanderso-nrel:
## Description

The sphinx build in #127 is failing because of #126, and I figured might as well fix #105 while we're at it.  

## Checklist

- [x] Partially addresses #105
- [x] Closes #126
- ~[ ] Added new API functions to `docs/api.rst`~
- ~[ ] Clearly documented all new API functions with [PEP257](https://www.python.org/dev/peps/pep-0257/) and [numpydoc](https://numpydoc.readthedocs.io/en/latest/format.html) compliant docstrings~
- ~[ ] Non-API functions clearly documented with docstrings or comments as necessary~
- ~[ ] Added tests to cover all new or modified code~

### DIFF
--- a/pvanalytics/tests/features/test_clearsky.py
+++ b/pvanalytics/tests/features/test_clearsky.py
@@ -1,10 +1,14 @@
 """Tests for feature labeling functions."""
 import pytest
 import pandas as pd
+import pvlib
 from pvanalytics.features import clearsky
 
+from pkg_resources import parse_version
+is_old_pvlib = parse_version(pvlib.__version__) < parse_version("0.9.0")
 
-@pytest.mark.skip(reason="GH #105")
+
+@pytest.mark.skipif(is_old_pvlib, reason="GH #105")
 @pytest.mark.filterwarnings("ignore:Support for multi-dimensional indexing")
 def test_reno_identical(quadratic):
     """Identical clearsky and measured irradiance all True"""
@@ -14,7 +18,7 @@ def test_reno_identical(quadratic):
     assert clearsky.reno(quadratic, quadratic).all()
 
 
-@pytest.mark.skip(reason="GH #105")
+@pytest.mark.skipif(is_old_pvlib, reason="GH #105")
 @pytest.mark.filterwarnings("ignore:Support for multi-dimensional indexing")
 @pytest.mark.filterwarnings("ignore:invalid value encountered in")
 def test_reno_begining_end(quadratic):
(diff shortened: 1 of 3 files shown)


pvanalytics/tests/features/test_clearsky.py
```diff
@@ -1,10 +1,14 @@
...
 import pandas as pd
+import pvlib
 from pvanalytics.features import clearsky
 
+from pkg_resources import parse_version
+is_old_pvlib = parse_version(pvlib.__version__) < parse_version("0.9.0")
 
-@pytest.mark.skip(reason="GH #105")
+
+@pytest.mark.skipif(is_old_pvlib, reason="GH #105")
```
kanderso-nrel, line 11: Is this sufficient for #105?  It's not clear to me what the outcome of that issue was.

cwhanse: I'm OK with marking this test to be skipped, as a way to get beyond it, but I think the #105 is still relevant. IIRC, the action here is to replace the expected result in the test. The test was created using pvlib v0.8.1, which had a bug corrected in v0.9.0 via [1242](https://github.com/pvlib/pvlib-python/pull/1242).

kanderso-nrel: What test values are there to update?  These tests don't seem to use particularly precise hardcoded values anywhere, unless I'm missing something.  

An aside: the commit message on 09e93a0 will automatically close #105, and I'm not sure how to prevent that now, so we'll have to manually reopen that issue if this is merged.

cwhanse (approved):

PR #163, opened 2022-09-21
Component sum functions

### LINKED ISSUE #157: Adding component sum function calculations to the irradiance module
Hey all, Josh Peterson from the University of Oregon approached me at PVPMC, interested in getting some of his functions based on published works integrated into PVAnalytics. One of the first functions he shared with me is based on the irradiance component sum equations (incredibly similar to what we have for our QCRad checks, but this directly calculates a field based on the other fields, with potential fields being GHI, DNI, or DHI). I'm sharing this function on his behalf to encourage further discussion. From what he's shared, I'd envision it as a way to calculate data fields if they aren't currently present in a data set for future analysis, as I think a lot of the validation of existing irradiance fields (say, comparing ground-based GHI measurements to component sum-calculated GHI using a GHI ratio) is already sitting up there in our QCRad checks. Let me know what you guys think about potentially adding this. I was going to flag an issue for each of the functions he'd like to add to encourage further discussion by the entire team. I can work with Josh to add the algorithms to the package, if we decide as a group to incorporate them.

Here is outline of the functionality, which is being used to estimate GHI using DNI and DHI values:

```
import numpy as np

def subroutineGHIcalc(dni, dhi, sza, szalimit, fillvalue, fillnightoption):
    '''
    Computes GHI from component sum equation
    ghi = dni * Cos(sza) + dhi

    Inputs: 
    dni is an array of floats
    dhi is an array of floats (Must be same units as DNI)
    sza is an array of floats (degrees)
    
    szalimit: float (degrees) SZA boundary between night and day. SZA values greater than the limit are filled with a constant
        default: 90
        
    fillvalue: float. The value that is used to fill in nighttime values.
        default: NA
    
    fillnightoption: 
        1: fill the nighttime value with the fill value (NA, 0, -99 etc)
        2: fill the nighttime value with the DHI value such that at night (GHI == DHI)
        Other: do nothing to the nighttimie values. compute them as they are. 

    Returns: GHI: array of floats, (same units as DNI)
    '''               
    # Not sure if we want to do a test to make sure the DNI, DHI, SZA are valid values.
    # Specifically are they floats. I don't think we would want to test if they are reasonable..
    # Compute the GHI value from the component sum equation
    ghi = dni * np.cos(sza * np.pi / 180) + dhi
    # Decide what you are going to do with the nighttime values
    if (fillnightoption == 1) |(fillnightoption == 2):
        # Find the locations where the sun is below the sza limit. 
        mask = (szalimit <= sza) 
    if (fillnightoption == 1):    
        # Replace the nighttime values with a fill value
        ghi[mask] = fillvalue
    elif fillnightoption == 2:
        # Replace the nighttime values with the DHI values. 
        # This will put 
        ghi[mask] =dhi[mask]
    return ghi

# A simple test
dni = np.array([1000,900,800])
dhi = np.array([50, 100, 200])
sza = np.array([100, 30, 30 ])

ghi = subroutineGHIcalc(dni, dhi, sza, 90, fillvalue=33, fillnightoption=1)

print(ghi)
```

kperrynrel:
Added logic for component sum equations for GHI, DHI, and DNI. Worked with @PetersonUOregon to build out the logic.

- [x] Closes #157 


cwhanse:
Two questions for discussion:
- would these functions (e.g., DNI = function of GHI, DHI) be better imported from pvlib-python?
- how do we protect against the inevitable "this value looks wrong" usually because of division by cos(zenith) when zenith is close to 90 (or >90)?

I like the `fill_nighttime` kwarg.

kperrynrel:
@cwhanse on your first point let me hook things up to pvlib--I like the idea of outsourcing here to existing packages because it makes our lives easier. I need to stew on point 2, maybe @PetersonUOregon may have some good insight.

PetersonUOregon:
The computation of DNI from GHI, DHI, and SZA is always a bit tricky near the horizon as @cwhanse said. 

As it currently is the function does have a variable szalimit that can be used to minimize some of this unwanted behavior. For example setting the szalimt to 85 instead of 90, would result in division of .09, (not zero). 

The other option here, is that we impose a QC to the output value to make sure that DNI does not exceed a clear sky model. But I expect that this is beyond the scope of this function.

kperrynrel:
@kanderso-nrel could you take a look at this PR? I couldn't find a good way to integrate any of the pvlib component sum functions into these functions, but the code is simplified enough that I'm not sure any integration is necessary. I also added an example gallery example.

kanderso-nrel:
I'll happily review this but I think we should make sure we're all on the same page about goals first. 

> would these functions (e.g., DNI = function of GHI, DHI) be better imported from pvlib-python?

I'm torn; "completing the irradiance triangle" could be viewed as a gap filling method more suited to pvanalytics or an irradiance prediction model more suited to pvlib.  I think I'd lean towards improving the functionality that pvlib already has along these lines (https://github.com/pvlib/pvanalytics/issues/157#issuecomment-1242538622), which as @kperrynrel points out is either incomplete (only capable of producing DNI) or not intended to be accessed by the user.  Maybe a good starting point is promoting the body of `ModelChain._complete_irradiance` into a new function `pvlib.irradiance.complete_irradiance`?

cwhanse:
To me, the two-to-three component calculations feel like they belong in pvlib, but the fill_nighttime function belongs here. I support "promoting the body of ModelChain._complete_irradiance into a new function pvlib.irradiance.complete_irradiance"

The code is in `pvlib.modelchain.ModelChain` because that's where it was being used, not for any good design reason.

kperrynrel:
Thanks @kanderso-nrel and @cwhanse --what do you guys think about me making two component sum functions for DHI and GHI calculations in PVlib, then importing these functions here, and applying the nighttime filter?

cwhanse:
Instead of adding new functions, I think we take the guts out of `pvlib.modelchain.ModelChain._complete_irradiance` and make a new function in `pvlib.irradiance`. I like that the current method handles all three cases.

kperrynrel:
@cwhanse just tagged a new issue in PVLib. I can work on it this week, and then once we go through a review process and PR in PVLib, we pick this back up? https://github.com/pvlib/pvlib-python/issues/1565

kanderso-nrel:
Now that https://github.com/pvlib/pvlib-python/issues/1567 has landed, what's the next step here?  If pvlib keeps to its new quarterly release schedule, the new `pvlib.irradiance.complete_irradiance` function won't be included in a pvlib release until December sometime.  I see two ways forward:

1. If we wait until the pvlib function is available to import, will we bump our minimum pvlib to 0.9.4 (currently [0.8.0](https://github.com/pvlib/pvanalytics/blob/master/setup.py#L37)), or maybe just require the newest pvlib for that one function? 
1. Alternatively, we could temporarily [vendor](https://stackoverflow.com/questions/26217488/what-is-vendoring) the pvlib function here with a TODO to eventually switch to importing it from pvlib.

cwhanse:
> 2. Alternatively, we could temporarily [vendor](https://stackoverflow.com/questions/26217488/what-is-vendoring) the pvlib function here with a TODO to eventually switch to importing it from pvlib.

I'm fine with this option.

wholmgren:
Make a new pvlib release! `git tag v0.9.4; git push upstream master` then merge the conda forge PR. forget about the rest. Or make a prerelease and pull that for CI purposes.

kperrynrel:
I definitely think @wholmgren 's suggestion is the easiest if we can just cut a new pre-release, but I also am not a pvlib maintainer so it's up to you guys on what you'd like to do! I'm going to get started on integrating the pvlib code into this PR today, and will just vendor the function over until we make a final decision.

wholmgren:
In case you do go that route: https://github.com/pvlib/pvlib-python/releases/tag/v0.9.4-alpha.1

pvanalytics/quality/irradiance.py
```diff
@@ -471,3 +471,174 @@ def daily_insolation_limits(irrad, clearsky, daily_min=0.4, daily_max=1.25):
...
+        series[mask] = fill_value
+    if fill_nighttime == 'equation':
+        # Use the nighttime equation GHI = 0 + DHI, which translates as:
+        # GHI_Calc (at night) = DHI_measured
+        # DHI_Calc (at night) = GHI_measured
+        # DNI_Calc (at night) = 0
+        if component == 'GHI':
+            series[mask] = component_sum_df['dhi'][mask]
+        elif component == 'DHI':
+            series[mask] = component_sum_df['ghi'][mask]
```
cwhanse, line 502: Are these reversed?

pvanalytics/quality/irradiance.py
```diff
@@ -471,3 +471,174 @@ def daily_insolation_limits(irrad, clearsky, daily_min=0.4, daily_max=1.25):
...
+        Pandas series of dni data, with datetime index. Must have the same
+        datetime index as ghi, dhi, and zenith series, when available.
+    dni_clear : Series, optional
+        Pandas series of clearsky dni data. Must have the same datetime index
+        as ghi, dhi, dni, and zenith series, when available. See
+        :py:func:`dni` for details.
+    zenith_limit: Float
+        Solar zenith boundary between night and day, in degrees.
+        Zenith values greater than the limit are filled with a
+        constant default of 90.
```
cwhanse, line 614:
```suggestion
        For calculation of the component sum, solar_zenith is set to 90 where
        solar_zenith > zenith_limit.
```

pvanalytics/quality/irradiance.py
```diff
@@ -471,3 +471,174 @@ def daily_insolation_limits(irrad, clearsky, daily_min=0.4, daily_max=1.25):
...
+    fill_value: Float, default np.nan
+        The value that is used to fill in nighttime value
+    fill_nighttime: String, default None
+        Options include 'fill_value', 'equation', None. This is the
```
cwhanse, line 618: This combination `fill_value` and `fill_nighttime` seems a bit redundant. `fill_nighttime='equation'` is equivalent to filling with DHI or GHI, so this behavior could be set by `fill_nighttime='GHI'` (or ='DHI'). `fill_nighttime='fill_value'` could be indicated if `fill_nighttime` is set to a numeric.

pvanalytics/quality/irradiance.py
kperrynrel: This is the intended implementation as the equation simplifies from:
GHI = DHI + DNI * cos(theta_z)
To:
 GHI = 0 + DHI
at nighttime (as DNI=0). By this logic, the GHI data is equal to the DHI data (and vice versa) at nighttime.

cwhanse: Ah, I understand. `component` is the Series to be filled.

pvanalytics/quality/irradiance.py
```diff
@@ -471,3 +471,174 @@ def daily_insolation_limits(irrad, clearsky, daily_min=0.4, daily_max=1.25):
...
+
+    Parameters
+    ----------
+    solar_zenith : Series
+        Zenith angles in decimal degrees, with datetime index.
+        Angles must be >=0 and <=180. Must have the same datetime index
+        as ghi, dhi, and dni series, when available.
+    ghi : Series, optional
+        Pandas series of dni data, with datetime index. Must have the same
+        datetime index as dni, dhi, and zenith series, when available.
```
cwhanse, line 600:
```suggestion
        Pandas series of GHI data, with datetime index. Must have the same
        datetime index as dni, dhi, and solar_zenith series, when available.
```

Similar edits needed for dhi, dni, and dni_clear.

pvanalytics/quality/irradiance.py
```diff
@@ -471,3 +471,174 @@ def daily_insolation_limits(irrad, clearsky, daily_min=0.4, daily_max=1.25):
...
+        Pandas series of the calculated values, based on the component sum
+        equation and corrected for nighttime periods.
+    '''
+    component_sum_df = _complete_irradiance(solar_zenith, ghi,
+                                            dhi, dni, dni_clear)
+    if ghi is None:
+        component = 'GHI'
+    elif dhi is None:
+        component = 'DHI'
+    else:
```
cwhanse, line 640: I'm thinking it would be better as 

```
elif dni is None:

else:
    raise error
```

kperrynrel: Updated to this in https://github.com/pvlib/pvanalytics/pull/163/commits/9f053f0fe38ab7c2b199f0b34349c2982bad4c36

pvanalytics/quality/irradiance.py
kperrynrel: @cwhanse I took a stab at addressing this in https://github.com/pvlib/pvanalytics/pull/163/commits/37eb655b50d66c9ca4cbbecc2cb339bc4be79bc0. Now we only have one variable fill_night_value, which can either be a float or int or 'equation'. Normally I don't like to do mixed types as a variable, but this does reduce the redundancy. Let me know what you think

pvanalytics/quality/irradiance.py
```diff
@@ -471,3 +471,177 @@ def daily_insolation_limits(irrad, clearsky, daily_min=0.4, daily_max=1.25):
...
+        Solar zenith boundary between night and day, in degrees.
+        For calculation of the component sum, solar_zenith is set to 90 where
+        solar_zenith > zenith_limit.
+    fill_night_value: String or float or int, default np.nan
+        Options include 'equation', float or int values (np.nan, 0, etc.).
+        This is the fill value for nighttime periods.
+        If a float or int value is passed (np.nan, 0 , -.5, etc.), then
+        nighttime values are filled using the fill_night_value parameter.
+        If 'equation' is used, nighttime fill periods are based on the
+        component sum equation for night (where DNI is 0):
```
cwhanse, line 620:
```suggestion
        If 'equation' is used, nighttime periods are filled using the
        component sum equation with DNI=0:
```

pvanalytics/quality/irradiance.py
```diff
@@ -471,3 +471,177 @@ def daily_insolation_limits(irrad, clearsky, daily_min=0.4, daily_max=1.25):
...
+            GHI = 0 + DHI
+        If neither of the above options are chosen, then the computed
+        nighttime values based of the component sum equation outpus are
+        returned.
```
cwhanse, line 624:
```suggestion
        If neither of the above options are chosen, then the computed
        nighttime values based of the component sum equation outpus are
        returned.
```
With `fill_night_value` default to `np.nan` this case appears to be excluded. The private function accepts anything which seems risky. What about `fill_night_value=None` being the default and the private function responds by filling with the component sum equation values (DNI being whatever it is)? If you agree, then add None as a permitted value.

I also think the private function should raise an Error if `fill_night_value` is not 'equation', float, int, np.nan or None.

kperrynrel:
```
    if isinstance(fill_night_value, float) | isinstance(fill_night_value, int):
        # Replace the nighttime values with a fill value--this can be np.nan,
        # which is a float
        series[mask] = fill_night_value
    elif fill_night_value == 'equation':
        # Use the nighttime equation GHI = 0 + DHI, which translates as:
        # GHI_Calc (at night) = DHI_measured
        # DHI_Calc (at night) = GHI_measured
        # DNI_Calc (at night) = 0
        if component == 'GHI':
            series[mask] = component_sum_df['dhi'][mask]
        elif component == 'DHI':
            series[mask] = component_sum_df['ghi'][mask]
        else:
            series[mask] = 0
    elif fill_night_value is None:
        pass
    else:
        raise ValueError("The fill_night_value variable must be None,"
                         " float or int, or 'equation'. Please change "
                         "the variable fill_night_value value.")
```

I reworked the logic to throw an error if np.nan (float), float/int, 'equation' or None is not passed. I changed the default setting to None, where we don't do anything, just pass the component-sum calculated series back.

pvanalytics/quality/irradiance.py
```diff
@@ -471,3 +471,185 @@ def daily_insolation_limits(irrad, clearsky, daily_min=0.4, daily_max=1.25):
...
+        None.
+        This is the fill value for nighttime periods.
+        If a float or int value is passed (np.nan, 0 , -.5, etc.), then
+        nighttime values are filled using the fill_night_value parameter.
+        If 'equation' is used, nighttime periods are filled using the
+        component sum equation with DNI=0:
+            GHI = 0 + DHI
+        If neither of the above options are chosen, then the computed
+        nighttime values based of the component sum equation outpus are
+        returned.
```
cwhanse, line 632:
```suggestion
        If None, then the nighttime values are based on the component sum
        equation.
```

pvanalytics/quality/irradiance.py
```diff
@@ -471,3 +471,185 @@ def daily_insolation_limits(irrad, clearsky, daily_min=0.4, daily_max=1.25):
...
+    the associated PVLib function can be directly leveraged.
+
+    Use the component sum equations to calculate the missing series, using
+    the other available time series. One of the three parameters (ghi, dhi,
+    dni) is passed as None, and the other associated series passed are used to
+    calculate the missing series value.
+    The "component sum" or "closure" equation relates the three
+    primary irradiance components as follows:
+    .. math::
+       GHI = DHI + DNI * cos(theta_z)
```
cwhanse, line 532:
```suggestion
    primary irradiance components as follows:

    .. math::
       GHI = DHI + DNI * cos(\theta_z)
```

pvanalytics/quality/irradiance.py
```diff
@@ -471,3 +471,185 @@ def daily_insolation_limits(irrad, clearsky, daily_min=0.4, daily_max=1.25):
...
+    dhi : Series, optional
+        Pandas series of dni data, with datetime index. Must have the same
+        datetime index as ghi, dni, and zenith series, when available.
+    dni : Series, optional
+        Pandas series of dni data, with datetime index. Must have the same
+        datetime index as ghi, dhi, and zenith series, when available.
+    dni_clear : Series, optional
+        Pandas series of clearsky dni data. Must have the same datetime index
+        as ghi, dhi, dni, and zenith series, when available. See
+        :py:func:`dni` for details.
```
cwhanse, line 551:
```suggestion
        as `ghi`, `dhi`, `dni`, and `solar_zenith` series, when available. See pvlib-python's 
        [dni](https://pvlib-python.readthedocs.io/en/stable/reference/generated/pvlib.irradiance.dni.html#pvlib.irradiance.dni) for details.
```

pvanalytics/quality/irradiance.py
```diff
@@ -471,3 +471,185 @@ def daily_insolation_limits(irrad, clearsky, daily_min=0.4, daily_max=1.25):
...
+    ----------
+    solar_zenith : Series
+        Zenith angles in decimal degrees, with datetime index.
+        Angles must be >=0 and <=180. Must have the same datetime index
+        as ghi, dhi, and dni series, when available.
+    ghi : Series, optional
+        Pandas series of GHI data, with datetime index. Must have the same
+        datetime index as dni, dhi, and solar_zenith series, when available.
+    dhi : Series, optional
+        Pandas series of dni data, with datetime index. Must have the same
```
cwhanse, line 608:
```suggestion
        Pandas series of DNI data, with datetime index. Must have the same
```

pvanalytics/quality/irradiance.py
```diff
@@ -471,3 +471,185 @@ def daily_insolation_limits(irrad, clearsky, daily_min=0.4, daily_max=1.25):
...
+
+    Parameters
+    ----------
+    solar_zenith : Series
+        Zenith angles in decimal degrees, with datetime index.
+        Angles must be >=0 and <=180. Must have the same datetime index
+        as ghi, dhi, and dni series, when available.
+    ghi : Series, optional
+        Pandas series of GHI data, with datetime index. Must have the same
+        datetime index as dni, dhi, and solar_zenith series, when available.
```
cwhanse, line 606:
```suggestion
        datetime index as `dni`, `dhi`, and `solar_zenith`, when available.
```

pvanalytics/quality/irradiance.py
```diff
@@ -471,3 +471,185 @@ def daily_insolation_limits(irrad, clearsky, daily_min=0.4, daily_max=1.25):
...
+    primary irradiance components as follows:
+    .. math::
+       GHI = DHI + DNI * cos(theta_z)
+
+    Parameters
+    ----------
+    solar_zenith : Series
+        Zenith angles in decimal degrees, with datetime index.
+        Angles must be >=0 and <=180. Must have the same datetime index
+        as ghi, dhi, and dni series, when available.
```
cwhanse, line 603:
```suggestion
        as `dni`, `dhi`, and `dni`, when available.
```

pvanalytics/quality/irradiance.py
```diff
@@ -471,3 +471,185 @@ def daily_insolation_limits(irrad, clearsky, daily_min=0.4, daily_max=1.25):
...
+    ghi : Series, optional
+        Pandas series of GHI data, with datetime index. Must have the same
+        datetime index as dni, dhi, and solar_zenith series, when available.
+    dhi : Series, optional
+        Pandas series of dni data, with datetime index. Must have the same
+        datetime index as ghi, dni, and zenith series, when available.
+    dni : Series, optional
+        Pandas series of dni data, with datetime index. Must have the same
+        datetime index as ghi, dhi, and zenith series, when available.
```
cwhanse, line 612:
```suggestion
        datetime index as `ghi`, `dhi`, and `solar_zenith`, when available.
```

pvanalytics/quality/irradiance.py
```diff
@@ -471,3 +471,185 @@ def daily_insolation_limits(irrad, clearsky, daily_min=0.4, daily_max=1.25):
...
+    dni_clear : Series, optional
+        Pandas series of clearsky dni data. Must have the same datetime index
+        as ghi, dhi, dni, and zenith series, when available. See
```
cwhanse, line 615:
```suggestion
        as `ghi`, `dhi`, `dni`, and `solar_zenith`, when available. See
```

pvanalytics/quality/irradiance.py
```diff
@@ -471,3 +471,185 @@ def daily_insolation_limits(irrad, clearsky, daily_min=0.4, daily_max=1.25):
...
+        :py:func:`dni` for details.
```
cwhanse, line 616:
```suggestion
        pvlib-python's 
        [dni](https://pvlib-python.readthedocs.io/en/stable/reference/generated/pvlib.irradiance.dni.html#pvlib.irradiance.dni) for details.
```

pvanalytics/quality/irradiance.py
```diff
@@ -471,3 +471,185 @@ def daily_insolation_limits(irrad, clearsky, daily_min=0.4, daily_max=1.25):
...
+    zenith_limit: Float
+        Solar zenith boundary between night and day, in degrees.
+        For calculation of the component sum, solar_zenith is set to 90 where
+        solar_zenith > zenith_limit.
```
cwhanse, line 620:
```suggestion
        For calculation of the component sum, `solar_zenith` is set to 90 where
        `solar_zenith > zenith_limit`.
```

cwhanse (approved):
Just a bit of final polishing. Thanks for your patience with this @kperrynrel

pvanalytics/quality/irradiance.py
kanderso-nrel: The backslash here might need the entire docstring to be prefixed with `r`, same as https://github.com/pvlib/pvlib-python/pull/1567#discussion_r996135976.  As a side note, I think it is better to have `\cos` instead of `cos` (renders as roman versus italics).

pvanalytics/quality/irradiance.py
kanderso-nrel: We have pvlib's docs linked via intersphinx, so I think ``` :py:func:`pvlib.irradiance.dni` ``` should work.  Although this docstring doesn't actually get rendered as HTML anywhere, so maybe @cwhanse's suggestion is better since this docstring will only ever be viewed as plaintext.

pvanalytics/quality/irradiance.py
```diff
@@ -471,3 +471,185 @@ def daily_insolation_limits(irrad, clearsky, daily_min=0.4, daily_max=1.25):
...
+    the other available time series. One of the three parameters (ghi, dhi,
+    dni) is passed as None, and the two series are used to
+    calculate the missing series. After calculation, the series is
+    run through a nighttime routine, where nighttime values are set based on
+    the fill_night_value parameter.
+
+    The "component sum" or "closure" equation relates the three
+    primary irradiance components as follows:
+    .. math::
+       GHI = DHI + DNI * cos(theta_z)
```
kanderso-nrel, line 596: Empty lines and backslashes should be applied here too

pvanalytics/quality/irradiance.py
kanderso-nrel: Same comment re: intersphinx

pvanalytics/quality/irradiance.py
```diff
@@ -471,3 +471,186 @@ def daily_insolation_limits(irrad, clearsky, daily_min=0.4, daily_max=1.25):
...
```
cwhanse, line 597:
```suggestion

    .. math::
   
       GHI = DHI + DNI * \cos(\theta_z)
```

I think this will fix the heading issue with the doc page.

pvanalytics/quality/irradiance.py
```diff
@@ -471,3 +471,188 @@ def daily_insolation_limits(irrad, clearsky, daily_min=0.4, daily_max=1.25):
...
+
+
+def calculate_component_sum_series(solar_zenith,
+                                   ghi=None,
+                                   dhi=None,
+                                   dni=None,
+                                   dni_clear=None,
+                                   zenith_limit=90,
+                                   fill_night_value=None):
+    '''
```
cwhanse, line 586:
```suggestion
    r'''
```

pvanalytics/quality/irradiance.py
```diff
@@ -471,3 +471,190 @@ def daily_insolation_limits(irrad, clearsky, daily_min=0.4, daily_max=1.25):
...
+        Pandas series of dni data, with datetime index. Must have the same
+        datetime index as ghi, dni, and zenith series, when available.
+    dni : Series, optional
+        Pandas series of dni data, with datetime index. Must have the same
+        datetime index as ghi, dhi, and zenith series, when available.
+    dni_clear : Series, optional
+        Pandas series of clearsky dni data. Must have the same datetime index
+        as `ghi`, `dhi`, `dni`, and `solar_zenith` series, when available. See pvlib-python's 
+        [dni](https://pvlib-python.readthedocs.io/en/stable/reference/generated/pvlib.irradiance.dni.html#pvlib.irradiance.dni) for details.
+    Returns
```
cwhanse, line 555:
```suggestion

    Returns
```

pvanalytics/quality/irradiance.py
```diff
@@ -471,3 +471,191 @@ def daily_insolation_limits(irrad, clearsky, daily_min=0.4, daily_max=1.25):
...
+    dni) is passed as None, and the two series are used to
+    calculate the missing series. After calculation, the series is
+    run through a nighttime routine, where nighttime values are set based on
+    the fill_night_value parameter.
+
+    The "component sum" or "closure" equation relates the three
+    primary irradiance components as follows:
+
+    .. math::
+
```
cwhanse, line 601:
```suggestion

```

pvanalytics/quality/irradiance.py
```diff
@@ -471,3 +471,190 @@ def daily_insolation_limits(irrad, clearsky, daily_min=0.4, daily_max=1.25):
...
+    run through a nighttime routine, where nighttime values are set based on
+    the fill_night_value parameter.
+
+    The "component sum" or "closure" equation relates the three
+    primary irradiance components as follows:
+
+    .. math::
+       GHI = DHI + DNI * \cos(\theta_z)
+
+    Parameters
```
cwhanse, line 603:
```suggestion

       GHI = DHI + DNI * \cos(\theta_z)


    Parameters
```

cwhanse:
@kperrynrel take a look at the function's doc page, the Parameters and Return sections aren't rendering right. I tried to fix it so I would understand what was wrong, but no luck. Reminder: you'll need to pull upstream since I was committing changes.

kperrynrel:
@cwhanse pulling your commits down now and seeing what I can do to debug. Thanks for taking a look

kanderso-nrel:
Looks OK to me on first glance, maybe your browser is caching an older broken version and a hard refresh will fix it?  The only rendering issue jumping out at me is the link to the pvlib docs; that's just due to using markdown link syntax instead of RST syntax.

kperrynrel:
image

Is this what you're seeing @cwhanse ? Mine also looks pretty screwy

kanderso-nrel:
I suspect the broken link syntax is to blame and the browser isn't finding enough characters it's willing to use to break the url across two lines.  Maybe a difference between chrome and firefox?  I see this on FF:

image

pvanalytics/quality/irradiance.py
```diff
@@ -471,3 +471,192 @@ def daily_insolation_limits(irrad, clearsky, daily_min=0.4, daily_max=1.25):
...
+        Pandas series of DNI data, with datetime index. Must have the same
+        datetime index as ghi, dni, and zenith series, when available.
+    dni : Series, optional
+        Pandas series of dni data, with datetime index. Must have the same
+        datetime index as `ghi`, `dhi`, and `solar_zenith`, when available.
+    dni_clear : Series, optional
+        Pandas series of clearsky dni data. Must have the same datetime index
+        as `ghi`, `dhi`, `dni`, and `solar_zenith`, when available. See
+        pvlib-python's 
+        [dni](https://pvlib-python.readthedocs.io/en/stable/reference/generated/pvlib.irradiance.dni.html#pvlib.irradiance.dni) for details.
```
kanderso-nrel, line 624:
```suggestion
        `dni <https://pvlib-python.readthedocs.io/en/stable/reference/generated/pvlib.irradiance.dni.html#pvlib.irradiance.dni>`_ for details.
```

cwhanse:
Looks OK to me now with latest commit.

kperrynrel:
Looks good to me as well @kanderso-nrel

PR #167, opened 2022-11-13
Poa qa

### LINKED ISSUE #123: QA methods for plane-of-array irradiance
[This paper](https://doi.org/10.1016/j.solener.2021.11.023) describes a workflow to establish quality of measured plane-of-array irradiance. The components and the workflow would be of significant value in this library.

abhisheksparikh:
- [ ] Closes #123 
- [ ] Added new API functions to `docs/api.rst`.
- [ ] Maintainer: Appropriate GitHub Labels and Milestone are assigned to the Pull Request and linked Issue.

This is a draft PR. I need help editing the `.rst` files.


pvanalytics/quality/irradiance.py
```diff
@@ -471,3 +471,133 @@ def daily_insolation_limits(irrad, clearsky, daily_min=0.4, daily_max=1.25):
...
+    References
+    ----------
+    .. [1] Elke Lorenz et. al, High resolution measurement network of global
+           horizontal and tilted solar irradiance in southern Germany with a
+           new quality control scheme, Solar Energy, Volume 231, 2022,
+           Pages 593-606, ISSN 0038-092X,
+           https://doi.org/10.1016/j.solener.2021.11.023.
+    """
+    # Defining the normal irradiance at the top of atmosphere in W/m^2
+    dni_extra = 1367
```
kanderso-nrel, line 561: Rather than hardcoding it, I suggest making this an optional parameter with default value 1367.0

pvanalytics/quality/irradiance.py
```diff
@@ -471,3 +471,133 @@ def daily_insolation_limits(irrad, clearsky, daily_min=0.4, daily_max=1.25):
...
+    aoi = aoi.clip(lower=0, upper=90)
+
+    # Determining the upper limit
+    upper_limit = 0.9 * dni_extra * (cosd(aoi))**1.2 + 300
+
+    # Setting upper limit as 0 when solar zenith is > 90 (night time)
+    upper_limit[solar_zenith > 90] = 0
+
+    # Setting upper limit as undefined where solar_zenith is not available
+    upper_limit[solar_zenith.isna()] = np.nan
```
kanderso-nrel, line 491: Should we do the same for `aoi.isna()`?

pvanalytics/quality/irradiance.py
```diff
@@ -471,3 +471,133 @@ def daily_insolation_limits(irrad, clearsky, daily_min=0.4, daily_max=1.25):
...
+
+    # Renaming upper_limit series to 'upper_limit'
+    upper_limit.rename('upper_limit')
+    return (upper_limit)
```
kanderso-nrel, line 495:
```suggestion
    return upper_limit
```
Just a style nitpick, and same for the other functions

pvanalytics/quality/irradiance.py
```diff
@@ -471,3 +471,133 @@ def daily_insolation_limits(irrad, clearsky, daily_min=0.4, daily_max=1.25):
...
+    upper_limit = 0.9 * dni_extra * (cosd(aoi))**1.2 + 300
+
+    # Setting upper limit as 0 when solar zenith is > 90 (night time)
+    upper_limit[solar_zenith > 90] = 0
+
+    # Setting upper limit as undefined where solar_zenith is not available
+    upper_limit[solar_zenith.isna()] = np.nan
+
+    # Renaming upper_limit series to 'upper_limit'
+    upper_limit.rename('upper_limit')
```
kanderso-nrel, line 494: I think `Series.rename` returns a renamed copy while leaving the original unchanged, so this line isn't doing anything useful.  I'd say we don't need a name on this Series anyway so might as well just delete (and same for the analogous line in the lower limit function).

pvanalytics/quality/irradiance.py
```diff
@@ -471,3 +471,133 @@ def daily_insolation_limits(irrad, clearsky, daily_min=0.4, daily_max=1.25):
...
+           Pages 593-606, ISSN 0038-092X,
+           https://doi.org/10.1016/j.solener.2021.11.023.
+    """
+    # Defining the normal irradiance at the top of atmosphere in W/m^2
+    dni_extra = 1367
+
+    # Making sure that the input are in series
+    poa_global = pd.Series(poa_global)
+    aoi = pd.Series(aoi)
+    solar_zenith = pd.Series(solar_zenith)
```
kanderso-nrel, line 566: If the docstring specifies Series for the inputs (which it does), no need for this IMHO

pvanalytics/quality/irradiance.py
```diff
@@ -471,3 +471,133 @@ def daily_insolation_limits(irrad, clearsky, daily_min=0.4, daily_max=1.25):
...
+
+    # Finding the upper and lower limit
+    upper_limit = _upper_poa_global_limit_lorenz(aoi, solar_zenith, dni_extra)
+    lower_limit = _lower_poa_global_limit_lorenz(solar_zenith, dni_extra)
+
+    # Initiating a poa_global_limit_int_flag series
+    poa_global_limit_int_flag = \
+        pd.Series(np.zeros(len(solar_zenith)).astype(np.int64),
+                  index=solar_zenith.index)
```
kanderso-nrel, line 575:
```suggestion
    poa_global_limit_int_flag = pd.Series(0, index=solar_zenith.index)
```
Passing an integer `0` results in `int64` dtype by default

pvanalytics/quality/irradiance.py
```diff
@@ -471,3 +471,133 @@ def daily_insolation_limits(irrad, clearsky, daily_min=0.4, daily_max=1.25):
...
+
+    # Initiating a poa_global_limit_bool_flag series
+    poa_global_limit_bool_flag = \
+        pd.Series(np.full(len(solar_zenith), True, dtype=bool),
+                  index=solar_zenith.index)
```
kanderso-nrel, line 580:
```suggestion
    poa_global_limit_bool_flag = pd.Series(True, index=solar_zenith.index)
```

pvanalytics/tests/quality/test_irradiance.py
```diff
@@ -314,3 +315,72 @@ def test_daily_insolation_limits_uneven(albuquerque):
         pd.Series(True, index=ghi.index),
         check_names=False
     )
+
+
+def test_check_poa_global_limits_lorenz():
+    """Testing the function 'test_check_poa_global_limits_lorenz'"""
+
+    test_file = DATA_DIR / "lorenz_poa_test_dataset.csv"
```
kanderso-nrel, line 323: Where did this dataset come from?

kanderso-nrel (commented):
Thanks @abhisheksparikh, this is looking pretty good.  This PR has some merge conflicts, largely due to editing the same files as #163 did.  If you need help resolving those let us know.

Also, the reference says:
> Additionally, we limit the maximal step between two consecutive irradiance measurements to 1000 W/m2 for both GHI and GTI as proposed for also in Espinar et al. (2011).

Do we need to do that as well in order to claim we have fully implemented the method as published?

pvanalytics/quality/irradiance.py
abhisheksparikh: Good idea!

pvanalytics/quality/irradiance.py
abhisheksparikh: Wow! I successfully managed to make it more complicated than it should have been.

pvanalytics/quality/irradiance.py
```diff
@@ -471,3 +471,133 @@ def daily_insolation_limits(irrad, clearsky, daily_min=0.4, daily_max=1.25):
...
     )
     return good_days.reindex(irrad.index, method='pad', fill_value=False)
+
+
+def _upper_poa_global_limit_lorenz(aoi, solar_zenith, dni_extra):
+    r"""Function to calculate the upper limit of GTI
+    """
+    # Changing aoi to 90 degrees when solar zenith is greater than 90 (sun
+    # below horizon) or aoi is greater than 90 (which is not realistic).
+    aoi = aoi.mask(solar_zenith > 90, 90)
```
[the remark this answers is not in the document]
abhisheksparikh: Yes, I think line 481 `aoi = aoi.mask(solar_zenith > 90, 90)` is redundant. Are we okay AOI is kept between 0 and 90? Line 484 will be NaN for AOI > 90. I agree with the fact that AOI can be greater than 90 (theoretically).

pvanalytics/quality/irradiance.py
```diff
@@ -471,3 +471,133 @@ def daily_insolation_limits(irrad, clearsky, daily_min=0.4, daily_max=1.25):
...
+    r"""Test for limits on POA global using the equations described in
+    Section 6.1 of [1]_
+
+    Criteria from [1] are used to determine physically plausible
+    lower and upper bounds. Each value is tested and a value passes if
+    value > lower bound and value < upper bound. Lower bounds are
+    constant for all tests. Upper bounds are calculated as
+
+    .. math::
+        ub = min + mult * dni\_extra * cos( solar\_zenith)^{exp}
```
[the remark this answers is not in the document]
abhisheksparikh: Whoops! Thanks for pointing that out.

pvanalytics/tests/quality/test_irradiance.py
abhisheksparikh: It's 5-psm data for 39.74 -105.17 (NREL's Golden campus). I got the POA values from SAM (ran a simulation with default system parameters).

pvanalytics/quality/irradiance.py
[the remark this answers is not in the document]
kanderso-nrel: > Are we okay AOI is kept between 0 and 90? Line 484 will be NaN for AOI > 90.

I'm not sure it's necessary to clip with `lower=0` (negative aoi should never happen in the first place), but I don't see any harm in it.  The reference doesn't seem to discuss the `aoi > 90` case (maybe I missed it), but imposing an upper limit of 90 seems appropriate to me.

pvanalytics/tests/quality/test_irradiance.py
kanderso-nrel: Instead of testing a "real" dataset, would it be easy/possible to craft a set of individual conditions where each combination tests a certain aspect of the function?  The `qcrad` tests are a good example of this: https://github.com/pvlib/pvanalytics/blob/main/pvanalytics/tests/quality/test_irradiance.py#L30

We're more likely to consider and explore all edge cases by assembling the inputs manually; nice clean PSM3 + SAM might not produce the exceptional conditions a QC routine is looking for.

abhisheksparikh: I agree - Actually, I didn't point that out earlier, but I did introduce some artificial issues in the data to explore different conditions. The following snapshot shows a part of the time series with the flagged time steps (The original data was all good):

image

kanderso-nrel: OK, I see.  Thanks!  I still think it would be more transparent to store the problematic data like how qcrad does (in the pytest file as a dozen or rows in a numpy array instead of an external CSV) but I won't harp on it.

abhisheksparikh: I see - I will mimic what the test function for qcrad does. 👍

pvanalytics/quality/irradiance.py
```diff
@@ -660,3 +660,288 @@ def calculate_component_sum_series(solar_zenith,
...
+    dni_extra : float
+        normal irradiance at the top of atmosphere in W/m^2
+
+    Returns
+    -------
+    poa_global_limit_bool_flag : Series
+        True for each value that is physically possible.
+    poa_global_limit_int_flag : Series
+        Series of integers representing the flag numbers described in the
+        literature [1]
```
cwhanse, line 736:
```suggestion
        literature. [1]_
```

pvanalytics/quality/irradiance.py
```diff
@@ -660,3 +660,288 @@ def calculate_component_sum_series(solar_zenith,
...
+    Parameters
+    ----------
+    poa_global : Series
+        Global tilted irradiance in :math:`W/m^2`
+    solar_zenith : Series
+        Solar zenith angle in degrees
+    aoi : Series
+        Direct normal irradiance in :math:`W/m^2`
+    dni_extra : float
+        normal irradiance at the top of atmosphere in W/m^2
```
cwhanse, line 728:
```suggestion
        normal irradiance at the top of atmosphere in :math:`W/m^2`
```

pvanalytics/quality/irradiance.py
```diff
@@ -660,3 +660,288 @@ def calculate_component_sum_series(solar_zenith,
...
+
+    return (lower_limit)
+
+
+def check_poa_global_limits_lorenz(poa_global, solar_zenith, aoi,
+                                   dni_extra=1367):
+    r"""Test for limits on POA global using the equations described in
+    Section 6.1 of [1]_
+
+    Criteria from [1] are used to determine physically plausible
```
cwhanse, line 710:
```suggestion
    Criteria from [1]_ are used to determine physically plausible
```

pvanalytics/quality/irradiance.py
```diff
@@ -660,3 +660,288 @@ def calculate_component_sum_series(solar_zenith,
...
+
+    Notes
+    -----
+    The upper limit for `poa_global` is set to 0 when `solar_zenith` is greater
+    than 90 degrees. Missing values of `poa_global`, `solar_zenith`
+    and/or `aoi` will result in a `False` flag.
+
+    References
+    ----------
+    .. [1] Elke Lorenz et. al, High resolution measurement network of global
```
cwhanse, line 746:
```suggestion
    .. [1] Elke Lorenz et al., High resolution measurement network of global
```

pvanalytics/quality/irradiance.py
```diff
@@ -660,3 +660,288 @@ def calculate_component_sum_series(solar_zenith,
...
+    # Setting upper limit as 0 when solar zenith is > 90 (night time)
+    upper_limit_flag3[solar_zenith > 90] = 0
+
+    # Setting upper limit as undefined where solar_zenith is not available
+    upper_limit_flag3[solar_zenith.isna()] = np.nan
+
+    return upper_limit_flag3
+
+
+def _lower_ghi_limit_lorenz(solar_zenith, dni_extra):
```
cwhanse, line 826: This function looks identical to `_lower_poa_global_limit_lorenz` if that's true, then I would have only one `_lower_limit_lorenz`

pvanalytics/quality/irradiance.py
```diff
@@ -660,3 +660,288 @@ def calculate_component_sum_series(solar_zenith,
...
+    lower_limit[solar_zenith.isna()] = np.nan
+
+    return (lower_limit)
+
+
+def check_ghi_limits_lorenz(ghi, solar_zenith, dni_extra=1367):
+    r"""Test for limits on global horizontal irradiance using the equations
+    described in Section 6.1 of [1]_
+
+    Criteria from [1] are used to determine physically plausible
```
cwhanse, line 847:
```suggestion
    Criteria from [1]_ are used to determine physically plausible
```

pvanalytics/quality/irradiance.py
```diff
@@ -660,3 +660,288 @@ def calculate_component_sum_series(solar_zenith,
...
+    dni_extra : float
+        normal irradiance at the top of atmosphere in :math:`W/m^2`
+
+    Returns
+    -------
+    ghi_limit_bool_flag : Series
+        True for each value that is physically possible.
+    ghi_limit_int_flag : Series
+        Series of integers representing the flag numbers described in the
+        literature [1]
```
cwhanse, line 883:
```suggestion
        literature. [1]_
```

pvanalytics/quality/irradiance.py
```diff
@@ -660,3 +660,288 @@ def calculate_component_sum_series(solar_zenith,
...
+
+    Notes
+    -----
+    The upper limit for `ghi` is set to 0 when `solar_zenith` is greater
+    than 90 degrees. Missing values of `ghi` and/or `solar_zenith` will result
+    in a `False` flag.
+
+    References
+    ----------
+    .. [1] Elke Lorenz et. al, High resolution measurement network of global
```
cwhanse, line 893:
```suggestion
    .. [1] Elke Lorenz et al., High resolution measurement network of global
```

pvanalytics/tests/quality/test_irradiance.py
```diff
@@ -411,3 +411,213 @@ def test_calculate_dni_component(generate_RMIS_irradiance_series):
...
+                                                                 solar_zenith,
+                                                                 dni_extra)
+
+    assert_series_equal(expected_upper_limit,
+                        test_upper_limit,
+                        check_names=False)
+
+
+def test__lower_poa_global_limit_lorenz(lorenz_test_data):
+    """Testing lower poa global limit defined by Lorenz et. al"""
```
cwhanse, line 517:
```suggestion
    """Testing lower poa global limit defined by Lorenz et al."""
```

pvanalytics/quality/irradiance.py
abhisheksparikh: Merged

pvanalytics/quality/irradiance.py
```diff
@@ -660,3 +660,271 @@ def calculate_component_sum_series(solar_zenith,
...
+    poa_global_limit_int_flag = poa_global_limit_int_flag.mask(
+        ((poa_global > upper_limit) |
+         (poa_global < lower_limit)),
+        3
+    )
+
+    # Changing the poa_global_flag to 3 when the step change in poa values is
+    # more than 1000 W/m2
+    poa_global_limit_int_flag = poa_global_limit_int_flag.mask(
+        (abs(poa_global - poa_global.shift(1)) > 1000),
```
kanderso-nrel, line 773:
```suggestion
        poa_global.diff().abs() > 1000,
```

Maybe marginally cleaner.  If you agree, can do the same in the GHI function

pvanalytics/quality/irradiance.py
```diff
@@ -660,3 +660,271 @@ def calculate_component_sum_series(solar_zenith,
...
+    Parameters
+    ----------
+    poa_global : Series
+        Global tilted irradiance in :math:`W/m^2`
+    solar_zenith : Series
+        Solar zenith angle in degrees
+    aoi : Series
+        Direct normal irradiance in :math:`W/m^2`
+    dni_extra : float
+        normal irradiance at the top of atmosphere in :math:`W/m^2`
```
kanderso-nrel, line 728:
```suggestion
    poa_global : Series
        Global tilted irradiance [W/m^2]
    solar_zenith : Series
        Solar zenith angle [degrees]
    aoi : Series
        angle of incidence [degrees]
    dni_extra : float, default 1367
        normal irradiance at the top of atmosphere [W/m^2]
```

pvanalytics/quality/irradiance.py
```diff
@@ -660,3 +660,271 @@ def calculate_component_sum_series(solar_zenith,
...
+         (upper_limit.isna()) |
+         (lower_limit.isna())),
+        1
+    )
+
+    # Changing the poa_global_limit_bool_flag depending on
+    # poa_global_limit_int_flag
+    poa_global_limit_bool_flag = poa_global_limit_bool_flag.mask(
+        cond=poa_global_limit_int_flag != 0,
+        other=False)
```
kanderso-nrel, line 789:
```suggestion
    poa_global_limit_bool_flag = poa_global_limit_int_flag != 0
```

Does that do the same thing?  If so, can get rid of the initialization at the top as well

pvanalytics/quality/irradiance.py
```diff
@@ -660,3 +660,271 @@ def calculate_component_sum_series(solar_zenith,
...
+        1.5 * dni\_extra * cos(solar\_zenith)^{1.2} + 100)
+
+    Parameters
+    ----------
+    ghi : Series
+        Global tilted irradiance in :math:`W/m^2`
+    solar_zenith : Series
+        Solar zenith angle in degrees
+    dni_extra : float
+        normal irradiance at the top of atmosphere in :math:`W/m^2`
```
kanderso-nrel, line 858:
```suggestion
    ghi : Series
        Global horizontal irradiance [W/m^2]
    solar_zenith : Series
        Solar zenith angle [degrees]
    dni_extra : float, default 1367
        normal irradiance at the top of atmosphere [W/m^2]
```

pvanalytics/quality/irradiance.py
```diff
@@ -660,3 +660,271 @@ def calculate_component_sum_series(solar_zenith,
...
+         (upper_limit_flag2.isna()) |
+         (upper_limit_flag3.isna()) |
+         (lower_limit.isna())),
+        1
+    )
+
+    # Changing the ghi_limit_bool_flag depending on ghi_limit_int_flag
+    ghi_limit_bool_flag = ghi_limit_bool_flag.mask(
+        cond=ghi_limit_int_flag != 0,
+        other=False)
```
kanderso-nrel, line 928: Same comment here as in the GTI function

pvanalytics/tests/quality/test_irradiance.py
```diff
@@ -411,3 +411,199 @@ def test_calculate_dni_component(generate_RMIS_irradiance_series):
...
+
+    # Expected upper limit
+    expected_upper_limit = data['lower_limit_ghi']
+
+    # Setting up inputs
+    dni_extra = 1367
+    solar_zenith = data['solar_zenith']
+
+    # Testing upper limit
+    test_upper_limit = irradiance._lower_limit_lorenz(solar_zenith,
```
kanderso-nrel, line 577: Several comments and variable names here should be `lower_limit`

pvanalytics/tests/quality/test_irradiance.py
```diff
@@ -411,3 +411,199 @@ def test_calculate_dni_component(generate_RMIS_irradiance_series):
...
     # Make sure that periods where sza>90 are equal equal to GHI values
     assert all(dni_series_equation[sza_series > 90].dropna() == 0)
+
+
+@pytest.fixture
+def lorenz_test_data():
+
+    data = pd.DataFrame(
+        columns=['ghi', 'poa_global', 'solar_zenith', 'azimuth', 'aoi',
+                 'lower_limit_poa', 'upper_limit_poa',
```
kanderso-nrel, line 421: I think the `lower_limit_poa` column isn't currently used

kanderso-nrel (commented):
Looking good @abhisheksparikh, a few polishing comments below

pvanalytics/data/lorenz_poa_test_dataset.csv
```diff
@@ -0,0 +1,577 @@
+Time stamp,ghi,poa_global,solar_zenith,azimuth,surface_tilt,aoi,poa_global_limit_bool_flag,poa_global_limit_int_flag,lower_limit,upper_limit
+1/1/2022 0:00,0,0,163.276,356.877,0,0,TRUE,0,0,0
```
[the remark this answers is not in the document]
abhisheksparikh: The only reason I left it in there was that the jupyter notebook (which I plan to convert into an example) uses this file. But yeah, I will remove the file for now.

pvanalytics/quality/irradiance.py
abhisheksparikh: Ah! Much cleaner! 👍

pvanalytics/quality/irradiance.py
abhisheksparikh: I copied the math type from `check_ghi_limits_qcrad` function. May be a silly question, but to avoid such future changes, what style can I refer to when documenting such functions?

pvanalytics/quality/irradiance.py
abhisheksparikh: Neat! A slight modification: 

`poa_global_limit_bool_flag = poa_global_limit_int_flag == 0`.

Will remove the initialization, and apply similar changes for GHI.

pvanalytics/quality/irradiance.py
kanderso-nrel: Yes, sorry that pvanalytics is inconsistent.  This is the same style we're using in pvlib (although older code in pvlib is also inconsistent).  Someday we will go through and standardize both packages on our current style.  It's mostly  [numpydoc](https://numpydoc.readthedocs.io/en/latest/format.html) but with some of our own conventions like this way of formatting units.  I don't think we've documented this style anywhere, but the docstring of [pvlib.temperature.prilliman](https://github.com/pvlib/pvlib-python/blob/v0.9.3/pvlib/temperature.py#L831) is a good example.

pvanalytics/tests/quality/test_irradiance.py
abhisheksparikh: Ah - that's a bit embarrassing. Thanks for pointing!

abhisheksparikh:
Hi @kanderso-nrel, should I remove my changes from 0.1.3 rst to 0.1.4 rst?

pvanalytics/quality/irradiance.py
```diff
@@ -660,3 +660,267 @@ def calculate_component_sum_series(solar_zenith,
...
+    """
+    # Finding the upper and lower limit
+    upper_limit = _upper_poa_global_limit_lorenz(aoi, solar_zenith, dni_extra)
+    lower_limit = _lower_limit_lorenz(solar_zenith, dni_extra)
+
+    # Initiating a poa_global_limit_int_flag series
+    poa_global_limit_int_flag = pd.Series(0, index=solar_zenith.index)
+
+    # Initiating a poa_global_limit_bool_flag series
+    poa_global_limit_bool_flag = pd.Series(True, index=solar_zenith.index)
```
kanderso-nrel, line 760: No need for this now that `poa_global_limit_bool_flag` is redefined on line 787

pvanalytics/quality/irradiance.py
```diff
@@ -660,3 +660,267 @@ def calculate_component_sum_series(solar_zenith,
...
+    return upper_limit_flag2
+
+
+def _upper_ghi_limit_lorenz_flag3(solar_zenith, dni_extra):
+    r"""Function to calculate the upper limit of ghi for Flag 3
+    """
+    # Determining the upper limit
+    upper_limit_flag3 = np.minimum(
+        pd.Series(1.2 * dni_extra, index=solar_zenith.index),
+        1.5 * dni_extra * (cosd(solar_zenith))**1.2 + 100)
```
kanderso-nrel, line 813:
```suggestion
        1.5 * dni_extra * cosd(solar_zenith))**1.2 + 100
```

The 100 should not be multiplied by the leading coefficients, right?

pvanalytics/quality/irradiance.py
```diff
@@ -660,3 +660,267 @@ def calculate_component_sum_series(solar_zenith,
...
+    upper_limit_flag3 = _upper_ghi_limit_lorenz_flag3(solar_zenith, dni_extra)
+
+    # Finding the lower limit for flag 3
+    lower_limit = _lower_limit_lorenz(solar_zenith, dni_extra)
+
+    # Initiating a ghi_limit_int_flag series
+    ghi_limit_int_flag = pd.Series(0, index=solar_zenith.index)
+
+    # Initiating a ghi_limit_bool_flag series
+    ghi_limit_bool_flag = pd.Series(True, index=solar_zenith.index)
```
kanderso-nrel, line 891: This can also be removed

pvanalytics/quality/irradiance.py
```diff
@@ -660,3 +660,267 @@ def calculate_component_sum_series(solar_zenith,
...
+    ghi_limit_int_flag = ghi_limit_int_flag.mask(
+        (ghi > upper_limit_flag3) |
+        (ghi < lower_limit),
+        3
+    )
+
+    # Changing the ghi_limit_int_flag to 3 when the step change in ghi values
+    # is more than 1000 W/m2
+    ghi_limit_int_flag = ghi_limit_int_flag.mask(
+        (abs(ghi - ghi.shift(1)) > 1000),
```
kanderso-nrel, line 910: `.diff().abs()` could be used here too

kanderso-nrel (approved):
@abhisheksparikh can you add a note to the new [0.1.4](https://github.com/pvlib/pvanalytics/blob/main/docs/whatsnew/0.1.4.rst) file?  A few other minor comments below, otherwise LGTM!  Another review from @cwhanse would be good.

pvanalytics/quality/irradiance.py
abhisheksparikh: Silly me!

abhisheksparikh: Whoops! It seems, my original equation was correct - the last `)` was not encapsulating 100 to be multiplied with the leading coefficients, but merely completing the `np.minimum` statement. 

For e.g. for SZA of 60 deg, I get a value of 992.53.

docs/api.rst
```diff
@@ -197,6 +213,11 @@ the quality check.
         Algorithm for Surface Radiation Measurements, The Open Atmospheric
         Science Journal 2, pp. 23-37, 2008.
 
+.. [2]  Elke Lorenz et. al, High resolution measurement network of global
```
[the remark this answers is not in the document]
abhisheksparikh: Yes :) Thanks for pointing that out!

pvanalytics/quality/irradiance.py
kanderso-nrel: D'oh!  Yes, my eyes tricked me.  In that case I suggest removing the extraneous parentheses around the cosine term (i.e. change `(cosd(solar_zenith))**1.2` to `cosd(solar_zenith)**1.2`) to perhaps prevent future readers from tripping over the same thing.

pvanalytics/quality/irradiance.py
```diff
@@ -660,3 +660,267 @@ def calculate_component_sum_series(solar_zenith,
...
+    described in Section 6.1 of [1]_
+
+    Criteria from [1]_ are used to determine physically plausible
+    lower, upper bounds and step change. Each value is tested and a value
+    passes if value > lower bound and value < upper bound. Also, steps with
+    change in magnitude of more than :math:`1000 W/m^{2}` are flagged. Lower
+    bounds are constant for all tests. As defined in the paper, there are
+    two values of upper bounds calculated:
+    (1) Rare values - Flag 2
+    (2) Extreme values - Flag 3
```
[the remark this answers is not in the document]
wholmgren: I think this will make the rendered text look like the rst.

wholmgren (commented):
This is great to see, thanks @abhisheksparikh! 

`cosd` is rather expensive, so I recommend that the internal functions also accept `cos_solar_zenith`. I suppose that could be done in a future PR too.

Note some of the comments on the high level POA function apply to GHI function too.

pvanalytics/quality/irradiance.py
```diff
@@ -660,3 +660,267 @@ def calculate_component_sum_series(solar_zenith,
...
+    # Setting lower limit as undefined where solar_zenith is not available
+    lower_limit[solar_zenith.isna()] = np.nan
+
+    return (lower_limit)
+
+
+def check_poa_global_limits_lorenz(poa_global, solar_zenith, aoi,
+                                   dni_extra=1367):
+    r"""Test for limits on POA global using the equations described in
+    Section 6.1 of [1]_
```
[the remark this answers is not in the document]
kanderso-nrel: The problem is that the first line of the docstring gets used as the description/blurb in the API listing, and if it includes a `[1]_`, that pointer gets naively reinterpreted in the context of the API listing page.  In `pvlib/pvlib-python` we don't have any references on the API listing page that it can point to, so we get build warnings about unknown references.  Here in `pvlib/pvanalytics` we do have some references on that page, so the footnote does link up with them, but the linking is wrong: `check_ghi_limits_lorenz`'s blurb links to the Long & Shi reference since that's what `[1]_` means on this page.

So yes the preference for not citing in the first sentence remains.  I think choice of theme is irrelevant; it's an autodoc/autosummary problem.

pvanalytics/quality/irradiance.py
```diff
@@ -660,3 +660,267 @@ def calculate_component_sum_series(solar_zenith,
...
+
+    return upper_limit
+
+
+def _lower_limit_lorenz(solar_zenith, dni_extra):
+    r"""Function to calculate the lower limit of poa_global and ghi
+    """
+    # Setting the lower_limit at 0.
+    lower_limit = pd.Series(np.zeros(len(solar_zenith)),
+                            index=solar_zenith.index)
```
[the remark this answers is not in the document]
abhisheksparikh: Wow! I really manage to make things more complicated than they should be.

abhisheksparikh:
How does one avoid this?

image

I need to refer this link a couple of times, but in doing so, I am ending up with this unclean format. (Having the same issue in the functions in `irradiance.py`

AdamRJensen:
As noted in the *Notes* section: "The upper limit for ghi is set to 0 when solar_zenith is greater than 90 degrees." Similarly, the lower limit is also 0 when the zenith angle is greater than 75 degrees. Thus, at nighttime, any value that is not exactly zero is flagged which I think is undesirable (irradiance measurements at nighttime typically have a slightly negative value, which is often accounted for by other QC methods). I don't think the authors meant for these checks to be applied for nighttime, but I suggest that we reach out and ask them.

Also, I don't like bundling in the max. step function. I could easily see myself using a custom mixture of upper and lower boundaries and setting other values for the max. step. Thus, I suggest having a separate max. step function for increased flexibility (the max step function was first proposed by Espinar et al. (2011) as noted in the paper).

kanderso-nrel:
> How does one avoid this?

There's no way that I'm aware of.  If you want a laugh, see what it does to poor [read_tmy3](https://pvlib-python.readthedocs.io/en/stable/reference/generated/pvlib.iotools.read_tmy3.html).  Anyway don't worry about it here.

abhisheksparikh:
> As noted in the _Notes_ section: "The upper limit for ghi is set to 0 when solar_zenith is greater than 90 degrees." Similarly, the lower limit is also 0 when the zenith angle is greater than 75 degrees. Thus, at nighttime, any value that is not exactly zero is flagged which I think is undesirable (irradiance measurements at nighttime typically have a slightly negative value, which is often accounted for by other QC methods). I don't think the authors meant for these checks to be applied for nighttime, but I suggest that we reach out and ask them.
> 
> Also, I don't like bundling in the max. step function. I could easily see myself using a custom mixture of upper and lower boundaries and setting other values for the max. step. Thus, I suggest having a separate max. step function for increased flexibility (the max step function was first proposed by Espinar et al. (2011) as noted in the paper).

On the first point: I agree with the fact that real irradiance measurements at nighttime rarely are 0. Does somebody watching this PR know any of the authors? If not, I don't mind reaching out to them. 

On the second point: Yes, I think it would be better to separate the max step function, along with keeping the step value as an input (with $1000 W/m^2$ as default). The only thing against this change is that it deviates from the paper's criteria, which groups lower, upper and max step together. A middle way (which still deviates from the paper's criteria) is that we can keep the max step value user-defined, which would allow one to circumvent that particular filter by just using a really high value. However, I am completely fine with separating out the filter as a separate function :relaxed:.

AdamRJensen:
I know Elke and can reach out. I'll cc you if you send your email to arajen@dtu.dk.

I can dig the idea of the user specified max step - smart! Having it (also) be a separate function might still be better, but I'm interested in other's people's opinions.

cwhanse:
I think that flagging low irradiance at nighttime is OK in the context of pvanalytics. The package provides other functions which label day/night periods, and each quality function returns a time series of Boolean for the critieria it checks. The temptation is to regard 'True' here as equivalent to a valid measurement - and that is not always the case, as the discussion brings out.

> irradiance measurements at nighttime typically have a slightly negative value

That is certainly true for thermopile instruments, but not for all instruments. I think a reference cell would report a small positive value.

AdamRJensen:
Elke Lorenz has written the following to me:
> For the naming:  Do you think it would be o.k. to name the functions “pvlive” instead of “Lorenz”? I think that using the name of the main author of a paper is what you typically do, but this was really team work and so I think “pvlive” would be more appropriate.

Thus, I suggest that we change the name as requested.

abhisheksparikh:
Sounds good!

Abhi

kandersolar:
Friendly ping @abhisheksparikh about this PR :)

docs/api.rst
```diff
@@ -197,6 +213,11 @@ the quality check.
         Algorithm for Surface Radiation Measurements, The Open Atmospheric
         Science Journal 2, pp. 23-37, 2008.
 
+.. [2]  Elke Lorenz et al., High resolution measurement network of global 
+        horizontal and tilted solar irradiance in southern Germany with a new
+        quality control scheme, Solar Energy, Volume 231, 2022, Pages 593-606,
+        ISSN 0038-092X, https://doi.org/10.1016/j.solener.2021.11.023.
```
AdamRJensen, line 219:
```suggestion
.. [2]  E. Lorenz et al., High resolution measurement network of global 
        horizontal and tilted solar irradiance in southern Germany with a new
        quality control scheme, Solar Energy, Volume 231, pp. 593-606, 2022.
        ISSN 0038-092X, https://doi.org/10.1016/j.solener.2021.11.023.
```

pvanalytics/quality/irradiance.py
```diff
@@ -592,3 +592,260 @@ def calculate_component_sum_series(solar_zenith,
...
+
+    # Setting lower limit as undefined where solar_zenith is not available
+    lower_limit[solar_zenith.isna()] = np.nan
+
+    return lower_limit
+
+
+def check_poa_global_limits_pvlive(poa_global, solar_zenith, aoi,
+                                   dni_extra=1367.):
+    r"""Test for limits on POA global with Lorenz algorithm.
```
AdamRJensen, line 636:
```suggestion
    r"""Test for limits on POA global with pvlive algorithm.
```

pvanalytics/quality/irradiance.py
```diff
@@ -592,3 +592,260 @@ def calculate_component_sum_series(solar_zenith,
...
+    upper_limit_flag3[solar_zenith > 90] = 0
+
+    # Setting upper limit as undefined where solar_zenith is not available
+    upper_limit_flag3[solar_zenith.isna()] = np.nan
+
+    return upper_limit_flag3
+
+
+def check_ghi_limits_pvlive(ghi, solar_zenith, dni_extra=1367.):
+    r"""Test for limits on GHI with Lorenz algorithm.
```
AdamRJensen, line 752:
```suggestion
    r"""Test for limits on GHI with pvlive algorithm.
```

pvanalytics/quality/irradiance.py
```diff
@@ -592,3 +592,260 @@ def calculate_component_sum_series(solar_zenith,
...
+    dni_extra : float, default 1367.
+        normal irradiance at the top of atmosphere [W/m^2]
+
+    Returns
+    -------
+    poa_global_limit_bool_flag : Series
+        True for each value that is physically possible.
+    poa_global_limit_int_flag : Series
+        Series of integers representing the flag numbers described in the
+        literature. [1]_
```
AdamRJensen, line 664:
```suggestion
        [1]_.
```

pvanalytics/quality/irradiance.py
```diff
@@ -592,3 +592,260 @@ def calculate_component_sum_series(solar_zenith,
...
+         (poa_global < lower_limit)),
+        3
+    )
+
+    # Changing the poa_global_flag to 3 when the step change in poa values is
+    # more than 1000 W/m2
+    poa_global_limit_int_flag = poa_global_limit_int_flag.mask(
+        poa_global.diff().abs() > 1000,
+        3
+    )
```
AdamRJensen, line 701: I think this deserves it's own function. 

The step change limit is fairly common:

> For 1-min data Shafer et al. (2000) recommended a step change limit of 800 W/m2, although Espinar (2011) suggested 1000 W/m2 for GHI. Lorenz et al. (2022) also suggested using the 1000 W/m2 threshold for tilted irradiance measurements.

AdamRJensen: It could admittedly be done in a separate pull request.

AdamRJensen (commented):
The code seems good, so well done!

My main concern is that it seems like an awful lot of code to implement the integer flags. I would prefer to just have the individual functions, but it's not a strong preference.

pvanalytics/quality/irradiance.py
abhisheksparikh: Good point. I think one good reason to have the step change limit in here is to stay true to the list of checks the paper suggests.

pvanalytics/quality/irradiance.py
```diff
@@ -592,3 +592,260 @@ def calculate_component_sum_series(solar_zenith,
...
+    poa_global_limit_bool_flag : Series
+        True for each value that is physically possible.
+    poa_global_limit_int_flag : Series
+        Series of integers representing the flag numbers described in the
+        literature. [1]_
+
+    Notes
+    -----
+    The upper limit for `poa_global` is set to 0 when `solar_zenith` is greater
+    than 90 degrees. Missing values of `poa_global`, `solar_zenith`
```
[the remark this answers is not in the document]
abhisheksparikh: Not having thought about the significance of nighttime data, I don't have a strong opinion on this. But if I were to think of irradiance as a part of operational analysis of a PV plant, I would like to have some distinction between daytime and nighttime data.

abhisheksparikh:
> The code seems good, so well done!
> 
> My main concern is that it seems like an awful lot of code to implement the integer flags. I would prefer to just have the individual functions, but it's not a strong preference.

Yeah, I agree, the code is definitely longer for what it's doing.

pvanalytics/quality/irradiance.py
cwhanse: we could have a helper function that returns the step threshold, and pass it a kwarg 'pvlive' to return 1000.

pvanalytics/quality/irradiance.py
```diff
@@ -592,3 +592,260 @@ def calculate_component_sum_series(solar_zenith,
...
+
+def check_ghi_limits_pvlive(ghi, solar_zenith, dni_extra=1367.):
+    r"""Test for limits on GHI with pvlive algorithm.
+
+    Criteria from [1]_ are used to determine physically plausible
+    lower, upper bounds and step change. Each value is tested and a value
+    passes if value > lower bound and value < upper bound. Also, steps with
+    change in magnitude of more than :math:`1000 W/m^{2}` are flagged. Lower
+    bounds are constant for all tests. As defined in the paper, there are
+    two values of upper bounds calculated:
```
cwhanse, line 759:
```suggestion
    The pvlive algorithm [1]_ defines a lower bound, two upper bounds (rare and extreme),
    and a limit on the change between successive GHI values. A value is physically plausible
    if:
    
     * the value is greater than the lower bound
     * the value is less than both the rare and extreme upper bounds
     * the step change to the next GHI value is less than the step change limit
        of :math:`1000 W/m^{2}` 
```

pvanalytics/quality/irradiance.py
```diff
@@ -592,3 +592,260 @@ def calculate_component_sum_series(solar_zenith,
...
+
+    .. math::
+        upper\_limit_{\mathbf{Flag\_2}} = 1.2 * dni\_extra * cos(solar\_zenith)
+        + 50
+
+    For Flag 3
+
+    .. math::
+        upper\_limit_{\mathbf{Flag\_3}} = min(1.2 * dni\_extra,
+        1.5 * dni\_extra * cos(solar\_zenith)^{1.2} + 100)
```
cwhanse, line 774:
```suggestion

    The upper limit for rare values is

    .. math::
        upper\_limit_{\mathbf{Flag\_2}} = 1.2 * dni\_extra * cos(solar\_zenith)
        + 50

    The upper limit for extreme values is

    .. math::
        upper\_limit_{\mathbf{Flag\_3}} = min(1.2 * dni\_extra,
        1.5 * dni\_extra * cos(solar\_zenith)^{1.2} + 100)
```

pvanalytics/quality/irradiance.py
```diff
@@ -592,3 +592,260 @@ def calculate_component_sum_series(solar_zenith,
...
+    dni_extra : float, default 1367
+        normal irradiance at the top of atmosphere [W/m^2]
+
+    Returns
+    -------
+    ghi_limit_bool_flag : Series
+        True for each value that is physically possible.
+    ghi_limit_int_flag : Series
+        Series of integers representing the flag numbers described in the
+        literature. [1]_
```
cwhanse, line 791:
```suggestion
        Series of integers representing the flag numbers described in the
        literature. [1]_:

        * 0 the value is physically plausible
        * 1 the value is missing
        * 2 the value exceeds the rare value upper limit
        * 3 the value is less than the lower limit, exceeds the extreme value upper limit,
          or has a step change greater than the step change limit. 
```

pvanalytics/quality/irradiance.py
```diff
@@ -592,3 +592,260 @@ def calculate_component_sum_series(solar_zenith,
...
+
+    Notes
+    -----
+    The upper limit for `ghi` is set to 0 when `solar_zenith` is greater
+    than 90 degrees. Missing values of `ghi` and/or `solar_zenith` will result
+    in a `False` flag. Also, [1]_ mentions that the proposed limits are for
+    silicon sensors with a tilt of 25°.
```
cwhanse, line 798:
```suggestion
    The upper limits for `ghi` are set to 0 at night when `solar_zenith` is greater
    than 90 degrees. Missing values of `ghi` and/or `solar_zenith` will result
    in a `False` plausible value flag. Also, [1]_ mentions that the proposed limits are for
    silicon sensors with a tilt of 25°.
```

cwhanse (approved):

pvanalytics/quality/irradiance.py
cwhanse: > However, I am completely fine with separating out the filter as a separate function ☺️.

As am I, let's do this and have a `step_limit` kwarg or something.

abhisheksparikh: > we could have a helper function that returns the step threshold, and pass it a kwarg 'pvlive' to return 1000.

Can somebody help me with this? After some reading, I think I understand kwargs, but have never written a function with them. Some more guidance on how the functions' design would be helpful.

AdamRJensen:
The authors have responded to our question concerning what to do for nighttime values:

> Regarding the night values: in our implementation we set all night values to zero and set a Flag with value 1 for all values changed. But actually, **the method is not made for night values**. So as you wrote, not checking is also an option which is maybe better for general applications. For some applications night values might be of interest, we actually do not use them anywhere.

So for the integer output, I think we should set nighttime values to True. But for the bool series returned, I suggest not setting any nighttime limits but rather setting a flag of True or False depending on the outcome of #191.

Additionally, the authors also wrote:

>One other thing I have to point out. While checking your implementation we found a mistake in the publication, which is very unfortunate. We write I0 as solar constant with a fixed value. But actually I0 should be the Normal extraterrestrial irradiance corrected from Sun-Earth distance, which has a yearly cycle. We use the formula proposed by Spencer 1971 with a simplification suggested in “Energy Meteorology, Lecture Notes from Detlev Heinemann, 2002 Formula 2.16”

@abhisheksparikh This should be relatively straightforward to correct. You can look at the existing tests which do something very similar.

abhisheksparikh:
> So for the integer output, I think we should set nighttime values to True. But for the bool series returned, I suggest not setting any nighttime limits but rather setting a flag of True or False depending on the outcome of https://github.com/pvlib/pvanalytics/issues/191.
@AdamRJensen - I am confused by this: 'So for the integer output, I think we should set nighttime values to True'.

I will change the Normal extraterrestrial irradiance based on the formula.

AdamRJensen:
> @AdamRJensen - I am confused by this: 'So for the integer output, I think we should set nighttime values to True'.

A typo on my behalf - the flag for the integer should of course be 1 as stated by the authors 😄

AdamRJensen:
@antondriesse I wanted to bring your attention to this QC check for GTI.

adriesse:
Thanks @AdamRJensen. We have read Lorenz 2022.

PR #186, opened 2023-02-21
Speed up `pvanalytics.features.daytime.power_or_irradiance`

kperrynrel:
This PR includes some updates to the day-night mask logic based on time shift algorithm testing/validation. Primarily:

1) Algorithm optimization. Vectorizing the _correct_edge_of_day_errors() logic so it runs considerably faster with the same outputs

2) Adding an additional logic step looking at the length of repeating values in the sequence (day/night). If number of repeats is less than a certain time period, then these cases are set as NaN and forward-filled. This logic really helps to clean up erroneous classifications to sunrise and sunset can be estimated more accurately. 


kperrynrel:
The first part of the PR addresses code refactoring of the following private function _correct_edge_of_day_errors(). Specifically, the following line of code was particularly slow due to the lambda function: 
`    day_length = night.groupby(night.cumsum()).transform(
        lambda x: len(x) * minutes_per_value
    )`

I changed this line of code to the following:
`    day_periods = (~night).astype(int)
    day_length = 1 + day_periods.groupby(
        night.cumsum()).transform('sum') * minutes_per_value`

In terms of speed up, I tested on the albuqerque location in the test_daytime unit tests and got the following speedup with identical results:

image

This is a 50X speed up time. I wound up adding this refactor in because the original code was significantly bottlenecking runtimes when I was performing the time shift valdiation work for PVRW.

kperrynrel:
In addition to the code speedup, I added some additional logic to clear up shorter mislabeled periods and ffill them. Particular day or night periods shorter than 30 minutes long are immediately nullified and forward filled. This step is just there to clean up mislabeling issues that I encountered when testing the time shift algorithm (the 30 minute threshold was empirically derived during testing, but is also settable, so technically people could not apply the logic at all if they set nullify_repeat_count to 0):  
image

kperrynrel:
@kandersolar I finally have time to start piping in all of the time shift work from earlier this year (I'm v late), but I think this one is ready. The pytest failures appear to be something not related to this PR, from what I can tell (it's ruptures failing on later versions of Python, and none of the edits I made involve ruptures). Is there something going on with the versioning in the test enviro/have you seen this before?

kandersolar:
@kperrynrel it's a change in recent scipy versions; you just got unlucky and were the first person to trigger tests since the change was made :)

I think #196 fixes it.  Can you take a look, and if it seems right to you, we can merge that and then merge main into this PR to get tests passing here as well?

kperrynrel:
@kandersolar got your new scipy update in here and all tests are passing, so I'd say good for review!

pvanalytics/features/daytime.py
```diff
@@ -212,6 +219,13 @@ def power_or_irradiance(series, outliers=None,
     night = ((low_value & low_diff)
              | (low_value & low_median)
              | (low_diff & low_median))
+    # Nullify cases where the classification lasts 30 minutes or less
+    night_duplicates = _run_lengths(night)
+    if nullify_repeat_count is None:
+        nullify_repeat_count = int(30 / minutes_per_value)
+    night.loc[night_duplicates <= nullify_repeat_count] = np.nan
```
kandersolar, line 226: Probably we should add test(s) to verify that this new parameter behaves as expected

kandersolar (commented):
Sorry I forgot about this @kperrynrel.  By the way, is this function described in a publication anywhere?

pvanalytics/features/daytime.py
```diff
@@ -174,6 +175,13 @@ def power_or_irradiance(series, outliers=None,
         The length of the rolling window used for calculating the
         median length of the day when correcting errors in the morning
         or afternoon. [days]
+    nullify_repeat_count : int, default None
+        The cutoff number of repeat subsequent readings in a day-night
+        masking sequence, where sequences containing a lower repeat value
+        count are nullified and forward-filled day-night mask sequence. This
+        is done specifically to prevent erroneously labeled periods.
+        If None is passed, the logic nullifies periods where the classification
+        lasts 30 minutes or less.
```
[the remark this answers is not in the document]
kperrynrel: This is really similar logic, and I ran some controlled testing with my labeled datasets and this nullify_repeat_count was essentially acting as a short-term pre-filter before the hours_min logic kicked in, finding extra cases that hours_min couldn't as it relies on a rolling filter that looks at surrounding days at the exact same time (which may be wrong if most of the rolling days in the filter are also incorrect--this is how nullify_repeat_count helps, is it isn't effected by surrounding days' values which may be wrong). Including this logic does slightly improve time shift detection results (as this logic is used in the pipeline), but it's by a very small amount:

With nullify_repeat_count logic:  2.593 minutes mean error across 231 time series with corrected time shift issues
Without logic: 2.651 minutes mean error across 231 time series with corrected time shift issues

Given that the logic only makes a small difference, I'm opting to table it for now, think about a better long-term solution that isn't just throwing more and more logic/rules at the code and junking it out. Let me know your thoughts @kandersolar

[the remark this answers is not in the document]
kandersolar: Good to have some hard numbers!  I was already having trouble following the logic, so if the improvement is only minor then keeping things simple for now sounds good to me.

docs/whatsnew/0.2.0.rst
```diff
@@ -15,6 +15,9 @@ Breaking Changes
 
 Enhancements
 ~~~~~~~~~~~~
+* Updated private function :py:func:`~pvanalytics.features.daytime._correct_edge_of_day_errors`
+  to be more performant by vectorization; the original logic was using a lambda call that was
+  slowing the function speed down considerably. (:pull:`186`)
```
kandersolar, line 20: Might be good to mention the relevant public function, since that's what most readers would care about. Also might be interesting to include a rough speed-up factor?

kperrynrel: Updated in https://github.com/pvlib/pvanalytics/pull/186/commits/d71fc15c3804b1129ca7ea1faf59607a65a935de

pvanalytics/features/daytime.py
kperrynrel: Removed so no longer needed

pvanalytics/features/daytime.py
```diff
@@ -76,9 +76,9 @@ def _correct_edge_of_day_errors(night, minutes_per_value,
     # the day/night boundary at one end of the day - sunrise or sunset
     # - was correctly marked, it will be replaced with the rolling
     # median for that minute).
-    day_length = night.groupby(night.cumsum()).transform(
-        lambda x: len(x) * minutes_per_value
-    )
+    day_periods = (~night).astype(int)
+    day_length = 1 + day_periods.groupby(
+        night.cumsum()).transform('sum') * minutes_per_value
```
kandersolar, line 81:
```suggestion
    day_length = (1 + day_periods.groupby(
        night.cumsum()).transform('sum')) * minutes_per_value
```

Should it be this?

kperrynrel: Oops, you're right. I double checked the old outputs to the two variations above, and the updated version is the same:

image

docs/whatsnew/0.2.0.rst
```diff
@@ -15,6 +15,9 @@ Breaking Changes
 
 Enhancements
 ~~~~~~~~~~~~
+* Updated private function :py:func:`~pvanalytics.features.daytime.power_or_irradiance`
```
kandersolar, line 18:
```suggestion
* Updated function :py:func:`~pvanalytics.features.daytime.power_or_irradiance`
```

PR #197, opened 2023-08-09
Update the time shift Ruptures function with the new logic that was tested/validated at PVRW 2023

kperrynrel:
This update makes the changes to the original function so it now matches the finalized logic that I used for validation in the PVRW 2023 poster (https://www.nrel.gov/docs/fy23osti/85699.pdf). Most notably, we use binary segmentation instead of Pelt here, which results in a massive speedup while still being performant.

- [ ] Added tests to cover all new or modified code.

### DIFF
--- a/pvanalytics/quality/time.py
+++ b/pvanalytics/quality/time.py
@@ -138,7 +138,7 @@ def shifts_ruptures(event_times, reference_times,
     # so the timezone is irrelevant.
     time_diff = \
         event_times.tz_localize(None) - reference_times.tz_localize(None)
-    break_points = ruptures.BinSeg(model='rbf',
+    break_points = ruptures.Binseg(model='rbf',
                                    min_size=period_min).fit_predict(
                                        signal=time_diff.values,
                                        pen=prediction_penalty
--- a/pvanalytics/quality/time.py
+++ b/pvanalytics/quality/time.py
@@ -138,7 +138,7 @@ def shifts_ruptures(event_times, reference_times,
     # so the timezone is irrelevant. Get the time difference in minutes.
     time_diff = \
         (event_times.tz_localize(None) -
-         reference_times.tz_localize(None)).dt.total_seconds() / 60
+         reference_times.tz_localize(None))
     # Get the index before removing NaN's
     time_diff_orig_index = time_diff.index
     # # Remove any outliers that may skew the results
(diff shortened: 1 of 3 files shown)


pvanalytics/quality/time.py
```diff
@@ -81,27 +66,32 @@ def shifts_ruptures(event_times, reference_times, period_min=2,
...
-    round_up_from : int, optional
-        The number of minutes greater than a multiple of `shift_min` for a
-        shift to be rounded up. If a shift is less than `round_up_from` then
-        it will be rounded towards 0. If not specified then the shift will
-        be rounded up from `shift_min // 2`. Using a larger value will
-        effectively make the shift detection more conservative as small
-        variations will tend to be rounded to zero. [minutes]
     prediction_penalty : int, default 13
         Penalty used in assessing change points.
         See :py:meth:`ruptures.detection.Pelt.predict` for more information.
```
kandersolar, line 81: I guess this cross-reference should be updated to something with Binseg instead of Pelt

pvanalytics/quality/time.py
```diff
@@ -81,27 +66,32 @@ def shifts_ruptures(event_times, reference_times, period_min=2,
...
+    zscore_cutoff : int, default 2
+        Z-score cutoff / maximum for filtering out outliers in each identified
+        segment found via changepount detection
+    bottom_quantile_threshold : float, default 0
+        Bottom quantile threshold for each time series segment identified via
+        changepoint detection. All data below this threshold is not considered
+        when determining the mean value for the segment, which is later
+        rounded to the nearest `period_min` value
+    top_quantile_threshold : float, default 0.5
```
kandersolar, line 90: Just curious... these default values exclude the top 50% of values within each segment, right?  I'm surprised that such an symmetrical window is a good default choice.  What's the reasoning for it?

kandersolar (approved):
Looks reasonable to me!  A couple minor comments.

pvanalytics/quality/time.py
kperrynrel: Updated to binseg from pelt in https://github.com/pvlib/pvanalytics/pull/197/commits/6da960aea9359c85da39ccfcbba71cea69e83cdc

pvanalytics/quality/time.py
kperrynrel: @kandersolar this was experimentally derived, and meant to filter out noise. The original code did a 2/3rd cutoff to make the code more conservative about time shifts (we only want to suggest a shift if we're absolutely certain, and we want to error on the side of caution). Essentially this parameter allows the user to decide how cautious they want to be when estimating time shifts.

PR #206, opened 2024-02-08
Allow tz-aware timstamps in `gaps.trim`

qnguyen345:
Updated quality.gaps.trim function for updated versions of Pandas 2.0

- [ ] Closes #xxx

### DIFF
--- a/pvanalytics/quality/gaps.py
+++ b/pvanalytics/quality/gaps.py
@@ -409,9 +409,11 @@ def trim(series, days=10):
     """
     start, end = start_stop_dates(series, days=days)
     mask = pd.Series(False, index=series.index)
-    
+
     if start:
-        mask.loc[pd.to_datetime(start.date()).tz_localize(series.index.tz):pd.to_datetime(end.date()).tz_localize(series.index.tz)] = True
+        mask.loc[pd.to_datetime(start.date()).tz_localize(series.index.tz):
+                 pd.to_datetime(end.date()).tz_localize(series.index.tz)] = \
+            True
     return mask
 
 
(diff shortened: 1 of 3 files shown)


pvanalytics/quality/gaps.py
```diff
@@ -409,8 +409,11 @@ def trim(series, days=10):
     """
     start, end = start_stop_dates(series, days=days)
     mask = pd.Series(False, index=series.index)
+
     if start:
-        mask.loc[start.date():end.date()] = True
+        mask.loc[pd.to_datetime(start.date()).tz_localize(series.index.tz):
+                 pd.to_datetime(end.date()).tz_localize(series.index.tz)] = \
+            True
```
kandersolar, line 416: The return values of `start_stop_dates` seem to always have timestamps at midnight.  With that in mind, do we need to do any conversion like `.date()` here?  Couldn't we just use `start` and `end` as-is?

kandersolar (commented):
Thanks @qnguyen345 for the PR!  One question below.

pvanalytics/quality/gaps.py
qnguyen345: You are right mask.loc[start:end] = True would also works since adding .date() would change start and end from tz-aware to tz-naive, which made the Pandas error. I edited the code and have pushed your suggestion.

kperrynrel:
@kandersolar we went through all of the examples and the QA notebooks and this was the only function throwing a pandas error (data shifts module didn't have issues). When this is merged in, we should be good to go with pandas>=2.0.0

kandersolar (approved):
Thanks @qnguyen345!

PR #214, opened 2024-10-09
Add outside_domain parameter to irradiance ratio QC tests

### LINKED ISSUE #191: Values outside QC test domain are incorrectly flagged as in error
I've found that the QC checks in pvanalytics flag values outside the testing domain as erroneously, which I strongly believe is against the purpose of the testing domain.

As an example, the [BSRN comparison and diffuse ratio tests](https://epic.awi.de/30083/1/BSRN_recommended_QC_tests_V2.pdf) are stated to only be performed for GHI>50 W/m^2. The reason for this threshold is that the uncertainty at lower irradiance is too high to perform the test reliably. **This is NOT saying that values lower than 50 W/m^2 should be flagged.**

If values lower than 50 W/m^2 should be flagged, it certainly does not make sense for the BSRN lower limits on the individual components to be -4 W/m^2. For example, the below data point which matches the closure equation perfectly (i.e., GHI = DHI+DNI*cos(Z)) is flagged as erroneously because it's below 50 W/m^2.

        qcrad_consistency_flag = check_irradiance_consistency_qcrad(
            solar_zenith=pd.Series(80),
            ghi=40,
            dhi=40,
            dni=0)

I have a series of QC checks that I would like to add, but this issue is a blocker for now.

This issue is related to https://github.com/pvlib/pvanalytics/issues/188#issuecomment-1481728592 and https://github.com/pvlib/pvanalytics/pull/167#issuecomment-1373484368

AdamRJensen:
- [x] Closes #191
- ~[ ] Added new API functions to `docs/api.rst`.~

This PR adds an optional `out_of_bounds` parameter to the irradiance QC tests which allows the user to specify whether to Fail (False) or pass (True) a test when the conditions are outside of the valid test domain.


AdamRJensen:
@cwhanse, @kandersolar, @kperrynrel Before completing this I'd appreciate a review of the parameter naming and implementation and whether you can support this for future irradiance QC tests

cwhanse:
The definition of `out_of_bounds` confuses me: "Whether a test is failed (False) or passes (True) when test conditions are not satisfied." If the test is not satisfied, and the function returns True, what is the point of doing the test?

If instead, you want the test to not apply the GHI and zenith bounds, then "skip_bounds=False" makes more sense to me.

AdamRJensen:
> The definition of `out_of_bounds` confuses me: "Whether a test is failed (False) or passes (True) when test conditions are not satisfied." If the test is not satisfied, and the function returns True, what is the point of doing the test?
> 
> If instead, you want the control the internal workings of the test, and not apply the GHI and zenith bounds, then "skip_bounds=False" makes more sense to me.

My earlier definition probably wasn't good. The following is probably better:

    outside_domain_value : default False
        Value to return when conditions are outside the test domain.

What I'm trying to achieve is that a test isn't considered failing when the test conditions aren't met (e.g., irradiance is below 50 W/m^2). @cwhanse does this make more sense?

cwhanse:
That makes sense now. "out_of_domain" works for me, "out_of_domain_value" seems like the "value" part is optional.

AdamRJensen:
@cwhanse @kandersolar This PR is ready for review.

Cliff suggested naming the new parameter ``out_of_domain``, whereas I have named it ``outside_domain`` - let me know if you have preferences.

pvanalytics/quality/irradiance.py
```diff
@@ -319,6 +325,10 @@ def check_irradiance_consistency_qcrad(solar_zenith, ghi, dhi, dni,
 
     Notes
     -----
+    The ``ghi_bounds`` are applied differently in the two tests. The components
+    tests applies the bounds to the sum of the diffuse and direct components,
+    whereas the diffuse ratio tests applies the bounds to the measured ``ghi``.
```
cwhanse, line 330:
```suggestion
    The QCRad algorithm checks that the input GHI is consistent with the component
    sum :math:`DNI \times \cos \( zenith \) + DHI` of input DNI and DHI, and that the
    ratio :math:`\frac{DHI}{GHI}` is reasonable.

    In these two parts, the ``ghi_bounds`` are applied differently. In the components
    test, the bounds are applied to the component sum of diffuse and direct
    irradiance, whereas in the diffuse ratio test the bounds are applied to the
    measured ``ghi``.
```

cwhanse (approved):
My browser GH is behaving strangely so if the suggested changes look wonky, its not me.

pvanalytics/quality/irradiance.py
```diff
@@ -309,6 +312,9 @@ def check_irradiance_consistency_qcrad(solar_zenith, ghi, dhi, dni,
         value is a dict with keys 'zenith_bounds', 'ghi_bounds', and
         'ratio_bounds' and value is an ordered pair [lower, upper]
         of float.
+    outside_domain : default False
+        Value to return when the tests are not applicable, i.e., when the
+        inputs fall outside the test domain.
```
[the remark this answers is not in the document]
AdamRJensen: I think @kandersolar was against listing the specifics. I somewhat agree, as "irradiance values outside the test domain" might be confused as to above/below the threshold. Not a strong feeling though.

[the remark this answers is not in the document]
cwhanse: OK. I'm OK with "data" instead of "inputs".

pvanalytics/quality/irradiance.py
```diff
@@ -269,21 +269,24 @@ def _get_bounds(bounds):
...
 
+    flag = within_domain & quality.util.check_limits(
+        ratio, lower_bound=ratio_lb, upper_bound=ratio_ub)
+
+    return flag, within_domain
+
 
 def check_irradiance_consistency_qcrad(solar_zenith, ghi, dhi, dni,
-                                       param=None):
+                                       param=None, outside_domain=False):
```
cwhanse, line 289: Add "r" to the line below so that the math equations render.

pvanalytics/quality/irradiance.py
```diff
@@ -269,21 +269,24 @@ def _get_bounds(bounds):
...
     """Check consistency of GHI, DHI and DNI using QCRad criteria.
```
AdamRJensen, line 290:
```suggestion
    r"""Check consistency of GHI, DHI and DNI using QCRad criteria.
```

pvanalytics/quality/irradiance.py
AdamRJensen, line 317:
```suggestion
        input data fall outside the test domain.
```

AdamRJensen:
@kandersolar a review of this would be appreciated

kandersolar (commented):
Should the `.. warning::` at the top of this function docstring (above `Parameters`) be edited, perhaps to mention this new parameter?

Tests look good to me.

pvanalytics/quality/irradiance.py
```diff
@@ -341,18 +356,30 @@ def check_irradiance_consistency_qcrad(solar_zenith, ghi, dhi, dni,
...
-                           bounds=bounds['high_zenith'])
-        | _check_irrad_ratio(ratio=dhi_ratio, ghi=ghi, sza=solar_zenith,
-                             bounds=bounds['low_zenith']))
+    flag_lz, within_domain_lz = _check_irrad_ratio(
+        ratio=dhi_ratio, ghi=ghi, sza=solar_zenith,
+        bounds=bounds['low_zenith'])
+    flag_hz, within_domain_hz = _check_irrad_ratio(
+        ratio=dhi_ratio, ghi=ghi, sza=solar_zenith,
+        bounds=bounds['high_zenith'])
+    within_domain_hz = within_domain_lz | within_domain_hz
```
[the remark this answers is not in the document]
AdamRJensen: This is indeed incorrect and left over from some refactoring/testing I did.

AdamRJensen, line 378:
```suggestion
```

pvanalytics/quality/irradiance.py
```diff
@@ -309,6 +312,9 @@ def check_irradiance_consistency_qcrad(solar_zenith, ghi, dhi, dni,
         value is a dict with keys 'zenith_bounds', 'ghi_bounds', and
         'ratio_bounds' and value is an ordered pair [lower, upper]
         of float.
+    outside_domain : bool, default False
```
AdamRJensen, line 315:
```suggestion
    outside_domain : bool, default False, optional
```
@kandersolar When labeling a variable as optional, do you still keep the default value?

kandersolar: We use `optional` when there is a default value, but the specific value is not really informative or useful to document.  So either document the default value, or call it `optional`, but not both.

> Also, the outside_domain variable need not be bool, it could really be anything, e.g., nan.

Good point.  What can we document for the type then?  `scalar`?

AdamRJensen: I have changed it to scalar

kandersolar:
> Should the .. warning:: at the top of this function docstring (above Parameters) be edited, perhaps to mention this new parameter?

@AdamRJensen any thoughts on this?  If you want to leave it the way it is, then LGTM and fine to merge

AdamRJensen:
This is ready to be merged. Thanks for the help

kandersolar (approved):

PR #219, opened 2025-04-16
Switch from `setup.py` to `pyproject.toml`

kandersolar:
This PR updates our packaging setup to store project metadata in pyproject.toml instead of setup.py, following PEP621.  We have done this with other packages already: https://github.com/pvlib/pvlib-python/pull/1910 and https://github.com/pvlib/twoaxistracking/pull/58

I checked the wheel contents and found them equivalent.

PR #220, opened 2025-04-16
Drop python 3.8

kandersolar:
It reached EOL some time ago and is now failing to install in the CI.  Drop it!


kandersolar:
Docs build failure is unrelated and will be fixed in #219.

kandersolar:
I assume these simple changes will not be controversial, so to unblock #219 I will merge this without review.

PR #221, opened 2025-04-16
Fix issue with `window_length` in `reno` clearsky detection

kandersolar:
- [ ] Closes #xxx
- [ ] Added tests to cover all new or modified code.
- [ ] Clearly documented all new API functions with [PEP257](https://www.python.org/dev/peps/pep-0257/) and [numpydoc](https://numpydoc.readthedocs.io/en/latest/format.html) compliant docstrings.
- [ ] Added new API functions to `docs/api.rst`.
- [ ] Non-API functions clearly documented with docstrings or comments as necessary.

`reno` was accidentally passing `window_length` to the `infer_limits` parameter.  That means the window length was not being provided to the underlying pvlib function and instead the default was used.  It also means that, since non-zero numbers are truthy, presumably this function has effectively had `infer_limits=True`.

@cwhanse should `infer_limits=True` be retained here?

### DIFF
--- a/docs/whatsnew/v0.2.3.rst
+++ b/docs/whatsnew/v0.2.3.rst
@@ -10,7 +10,8 @@ Enhancements
 
 Bug Fixes
 ~~~~~~~~~
-
+* :py:func:`pvanalytics.features.clearsky.reno` now correctly passes
+  ``window_length`` to the underlying pvlib function. (:pull:`221`)
 
 Requirements
 ~~~~~~~~~~~~
--- a/pvanalytics/features/clearsky.py
+++ b/pvanalytics/features/clearsky.py
@@ -72,7 +72,7 @@ def reno(ghi, ghi_clearsky):
         ghi,
         ghi_clearsky,
         ghi.index,
-        window_length,
+        window_length=window_length,
         lower_line_length=-5*scale_factor,
         upper_line_length=10*scale_factor,
         slope_dev=8*scale_factor


kandersolar:
python 3.8 CI failures should be ignored; we'll stop using 3.8 in #220.

cwhanse:
> should infer_limits=True be retained here?

I think `infer_limits=True` is better technically, but keeping that implied parameter would require fixing two other problems: 1) removing the explicit parameters calculated using `scale_factor` e.g. `lower_line_length` and 2) constraining window length to 30 minutes or less in this line:

```
window_length = np.minimum(10*delta_minutes, 60.0)
```

I vote we leave `infer_limits` out in this bug fix, and open an issue to advance `detect_clearsky` to default to `infer_limits=True`.

@kperrynrel your opinion? I mostly don't want to break any workflow for PVFleets.

kperrynrel:
@cwhanse @kandersolar  thanks for checking in with us on this. Just took a look at our workflows and we're using the clearsky call associated with the Location() class in pvlib, which doesn't appear to use pvlib's clearsky.detect_clearsky() function so no issues expected on our end for changing this logic up. I'd vote for a separate PR for adding infer_limits since it's a modification and this is a bug fix.

cwhanse:
@kperrynrel the pvlib `Location.get_clearsky` returns clearsky model GHI etc. for the location. Does PVFleets separate the PV system data by clear/nonclear condition? If so, how is that done?

kandersolar:
In it goes!

Issue #143, opened 2022-06-29
Comparing sensor measurements with external data to detect issues


williamhobbs:
**Question:**
Is identifying gradually worsening issues with sensors in-scope for PVAnalytics? For example, pyranometers with soiling or other sources of drift, or slowly detaching back of module temperature sensors:

https://user-images.githubusercontent.com/45701090/176544862-1b9c3052-998e-4d8b-995b-05571ebdf491.mp4

**Idea:**
If the answer is yes, it seems that using external solar resource/weather data plus `pvlib` functions to estimate POA, module temperature, etc., could provide a useful reference beyond what can be done with, e.g., clear sky models. And PVAnalytics seems like a good home for tools to compare two (or more) measurement "channels" and flag alarming deviations. This could be done to better QC historical datasets and/or to check for sensor issues that need correction in "real time" (e.g., as part of weekly or monthly plant/site maintenance work).

**On data sources:**
Some users may have access to commercial near real time satellite data, and ERA5 could be valuable to anyone (with as little as 5 days of lag and going back to at least 1979). NSRDB PSM3 could work if the typical 6-18 months of lag is acceptable. Measurements from nearby PV plants or weather stations could also be an option in some cases. The tools could be agnostic to the reference data source, although results may vary based on site and data source. 

Some relevant info on getting ERA5 or commercial satellite data into pvlib:
- Open `pvlib` pull requests with iotool functions for getting ERA5 and MERRA2 data: https://github.com/pvlib/pvlib-python/pull/1264 and https://github.com/pvlib/pvlib-python/pull/1274, .
- Discussion on getting ERA5 data: https://github.com/pvlib/pvlib-python/discussions/1484.
- Discussion on getting SolarAnywhere data: https://github.com/pvlib/pvlib-python/discussions/1310. 
- Issue on getting Solcast data: https://github.com/pvlib/pvlib-python/issues/1313

**Models/statistics to implement:**
I don't have much to add here right now, other than some general ideas like:
- look at "normal" historical deviations to set bounds outside of which deviations could be flagged as "abnormal"
- allow for possible "normal" seasonal biases
- start with regular statistics, although some machine learning might later prove useful

**What else has been done here:**
- https://github.com/pvlib/pvanalytics/issues/110 and https://github.com/pvlib/pvanalytics/issues/123 are relevant, but I don't think they include trending measurements against independent modeled values (or measurements for totally separate but still nearby sensors)
- Maybe lots of other things I haven't researched well...

[...]

PR #223, opened 2025-05-09
Adding an outlier check for dc capacity/power

qnguyen345:
~~- [ ] Closes #xxx~~
~~- [ ] Clearly documented all new API functions with [PEP257](https://www.python.org/dev/peps/pep-0257/) and [numpydoc](https://numpydoc.readthedocs.io/en/latest/format.html) compliant docstrings.~~
~~- [ ] Added new API functions to `docs/api.rst`.~~
- [ ] Maintainer: Appropriate GitHub Labels and Milestone are assigned to the Pull Request and linked Issue.

There can be days were the system is not producing the desired power output. We can measure the daily performance against a PVWatts model to determine those outlier days. We can model a system's expected dc capacity/ power output from PVWatts using the system metadata and nsrdb weather data. We can then compare the modeled daily time series to the real time series to get a percent difference. If the percent difference is over a certain threshold and is producing much less/more than is expected, we can flag that day as an anomaly.


pvanalytics/quality/outliers.py
```diff
@@ -124,3 +126,233 @@ def hampel(data, window=5, max_deviation=3.0, scale=None):
         kwargs=kwargs
     )
     return deviation > max_deviation * mad
+
+
+def run_pvwatts_data_checks(power_series, nsrdb_weather_df):
```
kperrynrel: Add underscore as this is a private method

pvanalytics/quality/outliers.py
```diff
@@ -124,3 +126,233 @@ def hampel(data, window=5, max_deviation=3.0, scale=None):
...
+    r"""Run the PVWatts model with NSRDB data across the time period as inputs.
+
+    Parameters
+    ----------
+    tilt : Float
+        Tilt angle of site in degrees.
+    azimuth : Float
+        Azimuth angle of site in degrees.
+    dc_capacity : Float
+        DC capacity of the site.
```
kperrynrel: data stream

pvanalytics/quality/outliers.py
```diff
@@ -124,3 +126,233 @@ def hampel(data, window=5, max_deviation=3.0, scale=None):
...
+    if not isinstance(nsrdb_weather_df.index, pd.DatetimeIndex):
+        raise TypeError('Weather dataframe must have a datetime index.')
+    # Get the frequency of weather data
+    nsrdb_weather_df = pd.infer_freq(nsrdb_weather_df.index)
+    # Resample power series to same frequency as weather data
+    power_series = power_series.resample(nsrdb_weather_df).mean()
+    return power_series
+
+
+def run_pvwatts_model(tilt, azimuth, dc_capacity, dc_inverter_limit,
```
kperrynrel: private method

pvanalytics/quality/outliers.py
```diff
@@ -124,3 +126,233 @@ def hampel(data, window=5, max_deviation=3.0, scale=None):
...
+    nsrdb_weather_df : Pandas dataframe with datetime index
+        A pandas dataframe containing NSRDB PSM3 weather 'Temperature', 'DHI',
+        'DNI', and 'Wind Speed' columns with datetime index. The minimum
+        and maximum datetime matches the minimum and maximum datetime of the
+        actual data.
+    pct_threshold : Float
+        Percent difference threshold for flagging data as anomalies.
+        Defaulted to 50.
+    dc_capacity : None or Float
+        DC capacity of the site. If the inverter dc capacity is not
```
kperrynrel: data stream instead of site

pvanalytics/quality/outliers.py
```diff
@@ -124,3 +126,233 @@ def hampel(data, window=5, max_deviation=3.0, scale=None):
...
+        known, set as None so that it can be calculated by finding the 95%
+        percentile of the passed time series.
+        Defaulted to None.
+
+    Returns
+    -------
+    master_df : Pandas dataframe with datetime index
```
kperrynrel: rename master_df as it's generic

kperrynrel: Return pandas series of percent difference, add new function to determine if anomalous where output is boolean with datetime index

cwhanse:
My reaction is that `run_pvwatts_model` doesn't belong as a function in pvanalytics. As the code is, the PVWatts model is hardwired into the data check function `get_anomalous_days` so I couldn't use a different performance model as input for this check - that reduces reusability.

`get_anomalous_days` ought to have a more specific name.

+1 to @kperrynrel's comments about the output of `get_anomalous_days`.

kperrynrel:
Hey @cwhanse, Quyen put this together on our end as this was a specific request from @williamhobbs. Southern wants to run an outlier check for "abnormal" daily behavior based on expected PVWatts output (they're using a lot of the PVAnalytics routines already). If you don't think it's a good fit, we could send him the code directly? Can you think of another open source repo where it may be more appropriate?

cwhanse:
Would the example be sufficient for @williamhobbs? The prepackaged PVWatts model could be a function in the example, although then it's not importable.



For identifying the outliers from a percent absolute difference in daily values, only `predicted` and `actual` are needed, and this function could accept lots of inputs: power, energy, temperature...

williamhobbs:
(I think this issue is almost 100% relevant to my comment below: https://github.com/pvlib/pvanalytics/issues/143.)

Here's my summary of our in-person conversation, @cwhanse. Hopefully this captures everything (with new sketches!):

We talked about a more general function/set of functions to flag deviations in a signal (like power or back of module temperature) from a reference, which could be from a physically adjacent piece of hardware (like inverter or Tbom sensor) or from a simulated signal, which I'm most interested in. It would be up to the user to provide the reference timeseries.

Anomalies could be flagged if the deviation (absolute value?) exceeds some time-based threshold, e.g., off by 20% for 1 hr or 10% for one day. The threshold could be a curve based on a function with one or two parameters, or maybe a piece-wise function based on a table. See the sketch below. 

image

There could a be possible second support function that you feed historical "good" data to and it returns the threshold curve at some confidence interval (e.g., 95% or 99% of historical deviations where below this curve). I could see this being very useful, otherwise there could be a lot of trial and error for users. 

image

I imagine these concepts already exists somewhere. My quick web-searching turns up network traffic anomaly detection, but it seems to be based only on past trends, not on an independent reference "expectation".

cwhanse:
@williamhobbs @kperrynrel @qnguyen345 

I propose we close this PR and replace with the following development goals:

1) add functions to `outlier.py` that label outliers in a timeseries based on deviations from a reference signal. The current functions in `outlier.py` find outliers using either the time series' marginal distribution (`zscore`, `tukey`) or sliding windows of the time series `hampel`). I don't know how to define an outlier from the deviation of signal from reference, but perhaps the existing outlier methods apply?

2) add functions to `outlier.py` to identify threshold curves in a time series. The geometry is intuitive, but the statistics can seem complicated and out of reach. Maybe there's a version of quantile regression that can be applied here.

williamhobbs:
@cwhanse - your proposal sounds good to me. 

Maybe the quantile regression in `statsmodels` (already required of pvanalytics, I think) can be used for a non-linear fit like this, https://stats.stackexchange.com/a/474426/375881, but where the x-axis is the time averaging window length.

kperrynrel:
@cwhanse @williamhobbs also good with closing this and reopening another PR with the newly recommended logic. Thanks!

PR #229, opened 2025-09-29
Switch to PyPI's Trusted Publishing 

kandersolar:
See https://github.com/pvlib/pvlib-python/issues/2511


.github/workflows/publish.yml
kandersolar, line 1: This file is just a copy/paste from pvlib's current workflow (with the needed edits for this repository, of course)

